PR #132, opened 2013-04-30
Make LinkedResource actually useful

poizan42:
LinkedResource currently isn't very useful. You cannot set the name in the file table except by specifying the actual path the hash is calculated from.
Unless you change your current working directory to the directory of the assembly, you have to specify the full path to the resource, resulting in that full path being recorded in the assembly - which is definitely never what you want, and breaks terribly if you ever try to deploy such an assembly.
Otherwise if you specify a path that doesn't exists when writing the file you end up with an empty hash. This makes sn.exe (the strong name tool) unable to recalculate the hash as no space is reserved in the assembly.

These changes attempts to fix that by letting the user explicitly specify the name in the file table independently of the file path and the resource name, as well as letting the user specify the hash to be delay calculated (fills in a zero hash) as well as explicitly specifying the hash.

### DIFF
--- a/Mono.Cecil/AssemblyReader.cs
+++ b/Mono.Cecil/AssemblyReader.cs
@@ -622,10 +622,25 @@ namespace Mono.Cecil {
 				} else if (implementation.TokenType == TokenType.File) {
 					var file_record = ReadFileRecord (implementation.RID);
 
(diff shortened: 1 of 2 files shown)


poizan42:
Bump?

poizan42:
Rebased and fixed indentation

PR #165, opened 2014-04-16
Allow raw instructions when constructing modules.

alexrp:
This is an initial proposal to allow emitting arbitrary stuff into method bodies. It currently doesn't support doing so when loading modules as I'm not sure what the right thing to do is in that case; there's not really a good way to know if a method body should be interpreted as CIL or not. Besides, this is mainly to implement features similar to `.byte` in ILAsm.

The idea here is that a raw instruction is just a bunch of bytes and, optionally, an operand that is encoded after those bytes. This should cater to all use cases, I think. The optional operand is there so that users don't have to worry about the low-level details of encoding metadata tokens and maintaining type/method/field refs when writing modules.

### DIFF
--- a/Mono.Cecil.Cil/CodeWriter.cs
+++ b/Mono.Cecil.Cil/CodeWriter.cs
@@ -166,7 +166,12 @@ namespace Mono.Cecil.Cil {
 
 			for (int i = 0; i < size; i++) {
 				var instruction = items [i];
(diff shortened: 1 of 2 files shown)


jbevain:
I'm afraid that doesn't really make sense as is, this would require a global token fixup, as there's no way to know the tokens before the module is written.

PR #192, opened 2014-12-09
Add Resolve() to MemberReference

lambdaknight:
Add a pseudo-covariant version of Resolve() to MemberReference to create a more complete common interface on MemberReference. The Resolve() in MemberReference returns a IMemberDefinition and is implemented by ResolveImpl. Each subclass can override ResolveImpl to change the behavior. In addition, each subclass implements a new method which returns the specific *Definition type and hides the Resolve() method in MemberReference. If you call Resolve() as a {Event,Field,Method,Property,Type}Reference, you get the appropriate type, but if you call it as a MemberReference, you get a IMemberDefinition.

### DIFF
--- a/Mono.Cecil/EventDefinition.cs
+++ b/Mono.Cecil/EventDefinition.cs
@@ -165,7 +165,7 @@ namespace Mono.Cecil {
 			}
 		}
 
-		public override EventDefinition Resolve ()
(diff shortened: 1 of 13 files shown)


jbevain:
Fixed in 452294087c66ea61391892eab2dabe4543d7dd0f.
Thanks!

PR #194, opened 2015-01-08
Adding a new abstract level.

vik-borisov:
I added a new abstract layer. It is now possible to write external or internal unit tests


jbevain:
Hi,

Thanks for your contribution. Unfortunately I don't have plans to add an intermediate abstract layer to the Cecil API. One of the strength of the current API is its straightforwardness.

PR #197, opened 2015-01-15
Fix reading custom metadata so it doesn't blow up in Roslyn

shana:
Cecil blows up in VS 2015 because Roslyn :P

### DIFF
--- a/symbols/pdb/Microsoft.Cci.Pdb/PdbFunction.cs
+++ b/symbols/pdb/Microsoft.Cci.Pdb/PdbFunction.cs
@@ -342,21 +342,20 @@ namespace Microsoft.Cci.Pdb {
       int savedPosition = bits.Position;
       byte version;
       bits.ReadUInt8(out version);
-      if (version != 4) {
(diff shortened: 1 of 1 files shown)


jbevain:
Hey! Thanks! Do you happen to have a small binary which reproduces this for the test suite?

shana:
I can't release the one it blows up on, but I'm trying to repro on a separate project. Should I commit it to Test/Resources/assemblies when I have it?

jbevain:
symbols/pdb/Test/Resources/assemblies, yeah.

I've been unable to repro the exception you're preventing. All the pdb I've tried (created with VS2015's Roslyn compiler) emit a pdb version of 4 (but they add different kinds, we prevent it with d33d51499a0c19238f41db5a06ab01b1f7aae210).

shana:
You know what, nevermind, the tool that was calling cecil and blowing up was shipping a broken binary with the same version as the latest one, which made my tooling very confused. The latest cecil works just fine, sorry for the noise :)

jbevain:
:bow:

Issue #196, opened 2015-01-14
Update Mono.CompilerServices.SymbolWriter


DavidKarlas:
Upstream Mono.CompilerServices.SymbolWriter supports [Column, EndLine and EndColumn](https://github.com/mono/mono/blob/97dc571fe2481ec429c1ab537dd7bf6b9316f5aa/mcs/class/Mono.CompilerServices.SymbolWriter/MonoSymbolTable.cs#L186-L187).

jbevain:
That looks sweet, thanks! All the tests are passing?

DavidKarlas:
Yes. Tests are passing.

jbevain:
Fixed. Thanks!

PR #198, opened 2015-01-15
Sync Mono.CompilerServices.SymbolWriter with upstream for Column support

DavidKarlas:
All files in Mono.CompilerServices.SymbolWriter are copy paste(with removed trailing white-spaces).
MdbReader.cs is modified same as in https://github.com/mono/mono/commit/a1fc1ea0590d9e9ee6a4c00b43f723cc1bf81e7b


DavidKarlas:
This is fixing #196

DavidKarlas:
@jbevain I though AppVeyor does that :)... Anyway... I just ran unit tests and all passed.

DavidKarlas:
Any progress on this?(I don't want to be rude, but I can't push debugger-lib fixes until debugger-libs users have option to bump Mono.Cecil)...

jbevain:
@DavidKarlas I see, I'll have a look promptly. One thing I think was missing, was populating the Cecil structures with the new information if available, and serializing it back when writing.

jbevain:
The change to a Stream based reader also allow the implementation of the symbol reading from a stream.

DavidKarlas:
Yes, but didn't want to introduced changes... Also not sure what to throw on Guid missmatch: 
```csharp
throw new MonoSymbolFileException ("Symbol file `{0}' does not match assembly", mdbFilename);
```
Maybe add casting(as) stream to FileStream and using .File, but if casting fails(null) throw generic:
```csharp
throw new MonoSymbolFileException ("Symbol stream does not match assembly", mdbFilename);
```

jbevain:
@DavidKarlas yeah that sounds good.

DavidKarlas:
I didn't do it inside Mono.CompilerServices.SymbolWriter because I want to keep it in sync with upstream...

DavidKarlas:
Fixed formatting and here is response in case you didnt' see:
Not my fault someone moved configurations to .settings and made me use tool unaware of formatting policies...

jbevain:
Thanks a bunch! I think we're only missing support for reading/writing the new data from the MdbReader and Writer and this will be good to go.

I'm afraid I don't understand your previous answer. What's wrong with the current setup and configurations?

DavidKarlas:
I have to use VS which don't format code depending on formatting policy... and MD can't handle configurations in .settings...

DavidKarlas:
Friendly warning... debugger-libs were update to use .Column property

jbevain:
Hi,

I ended up merging this and addressing https://github.com/jbevain/cecil/pull/198#issuecomment-70629324 in 7826271ad97c5120c74bce30b7cc911464447475.

Thanks!

PR #205, opened 2015-02-25
fix importing owner-less Var/MVar types

tailsu:
Fix the crash when `new GenericParameter(position, type, module)`, which creates an ownerless generic parameter, is then `module.Import()`'ed (or any type reference using referencing it internally).

### DIFF
--- a/Mono.Cecil/Import.cs
+++ b/Mono.Cecil/Import.cs
@@ -587,10 +587,14 @@ namespace Mono.Cecil {
 				return imported_instance;
 			case ElementType.Var:
 				var var_parameter = (GenericParameter) type;
-				return context.TypeParameter (type.DeclaringType.FullName, var_parameter.Position);
+				return type.DeclaringType != null
+					? context.TypeParameter(type.DeclaringType.FullName, var_parameter.Position)
+					: type;
 			case ElementType.MVar:
 				var mvar_parameter = (GenericParameter) type;
-				return context.MethodParameter (mvar_parameter.DeclaringMethod.Name, mvar_parameter.Position);
+				return mvar_parameter.DeclaringMethod != null
+					? context.MethodParameter(mvar_parameter.DeclaringMethod.Name, mvar_parameter.Position)
+					: type;
 			}
 
 			throw new NotSupportedException (type.etype.ToString ());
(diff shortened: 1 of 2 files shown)


jbevain:
That looks good, thanks! Any chance you can add a Import test so we don't regress here?

tailsu:
You got it :)

tailsu:
ping?

jbevain:
Hey @tailsu,

Sorry for the delay processing this.

I've had a look at the test case, and it makes me wonder if the fix is correct.

The reason is that you're trying to import a reference to a (open) generic instantiation, instantiated with a generic argument that is a generic parameter that has no parent. Which shouldn't be legal.

There's no real reason for this to happen, and if we shouldn't throw a pesky NRE, we shouldn't fallback to importing a reference that doesn't make sense in the Cecil type system.

Now I'm curious as to what triggered this. Is this actual code that you're running?

tailsu:
It's close to my actual code. In my actual code I emit a `ldtoken` with this type (essentially `ldtoken IEnumerable<!0>`) in a method in a type with a generic argument.  My understanding is that owner-less `GenericParameter` instances are used to refer to the generic arguments in context (`!0`, `!!0`, etc.).

jbevain:
Nope :) We shouldn't have a GenericParameter without a DeclaringType or a DeclaringMethod.

It depends what you try to have. In your case, if you want an open instantiation of IEnumerable`1 you could write:

```csharp
	var openEnumerable = module.Import (typeof (IEnumerable<>));
	var instanceEnumerable = new GenericInstanceType (openEnumerable);
	instanceEnumerable.GenericArguments.Add (openEnumerable.GenericParameters [0]);
```

Makes sense?

tailsu:
I have:
```
public class C<T>
{
public RuntimeTypeHandle M()
{
return typeof(IEnumerable<T>).RuntimeMethodHandle;
}
}
```
I implement the return using `ldtoken class IEnumerable<!0>`, where the metadata signature of the type is `GenericInst Class IEnumerable`1 Var!0`. In this situation it makes no sense to have a named generic parameter, because Var!0 makes sense _in context_. It also works correctly at runtime and produces no PEVerify errors.

I construct the type using
```
var openEnumerable = module.Import (typeof (IEnumerable<>));
var instanceEnumerable = new GenericInstanceType (openEnumerable);
var T = new GenericParameter (0, genericParameterType, module); // Var!0
instanceEnumerable.GenericArguments.Add (T);
```

Now, from Cecil's standpoint, I could take the first generic parameter of C<T> and add that to `instanceEnumerable.GenericArguments`, but either way should be equally valid, I think.

jbevain:
For this, yes, you need to take T from C, as we maintain a coherent object model where you can expect to know who declared a GenericParameter.

We probably shouldn't expose this constructor :)

tailsu:
Bummer...

jbevain:
Sorry, thanks for the PR though :)

PR #214, opened 2015-05-11
Add support for treating System.Runtime as the core library

jbevain:


jbevain:
@dgrunwald works for you?

dgrunwald:
Yes, this works for ILSpy.

PR #217, opened 2015-05-26
Allow upstream consumers to override build settings.

### LINKED ISSUE #216: Allow upstream consumers to override build settings.
Projects consuming and distributing Cecil in environments they
don't control (i.e. an addin inside VS or XS) risk having the
environment load an incompatible version of Cecil from another
addin or location (depending on the process/domain configuration
and the probing paths set up for the host).

To shield from this issue, XS has forked Cecil, changed its
assembly name and signing key, and now it has to maintain that
fork in sync with upstream. Others with similar requirements
(in particular, the Xamarin Visual Studio extension team now)
will need to do the same, with the added maintainability burden.

This commit provides a conditional import that allows products
that distribute Cecil to tweak settings that will affect its
build, without having to fork the project unnecessarily. The
Mono.Cecil.overrides file is a plain MSBuild file with additional
properties that can exist anywhere upwards from the current Cecil
repo clone path, making it very flexible even if the project
is submoduled.

It's imported before the C# targets so that changes to the
assembly name, signing key, etc. are properly overwritten before
the C# targets process them.

kzu:
Projects consuming and distributing Cecil in environments they
don't control (i.e. an addin inside VS or XS) risk having the
environment load an incompatible version of Cecil from another
addin or location (depending on the process/domain configuration
and the probing paths set up for the host).

To shield from this issue, XS has forked Cecil, changed its
assembly name and signing key, and now it has to maintain that
fork in sync with upstream. Others with similar requirements
(in particular, the Xamarin Visual Studio extension team now)
will need to do the same, with the added maintainability burden.

This commit provides a conditional import that allows products
that distribute Cecil to tweak settings that will affect its
build, without having to fork the project unnecessarily. The
Mono.Cecil.overrides file is a plain MSBuild file with additional
properties that can exist anywhere upwards from the current Cecil
repo clone path, making it very flexible even if the project
is submoduled.

It's imported before the C# targets so that changes to the
assembly name, signing key, etc. are properly overwritten before
the C# targets process them.

Fixes #216

### DIFF
--- a/Mono.Cecil.csproj
+++ b/Mono.Cecil.csproj
@@ -1,6 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="4.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Import Project="Mono.Cecil.settings" />
   <PropertyGroup>
     <ProjectGuid>{D68133BD-1E63-496E-9EDE-4FBDBF77B486}</ProjectGuid>
     <RootNamespace>Mono.Cecil</RootNamespace>
@@ -139,5 +138,6 @@
     <Compile Include="Mono\Empty.cs" />
     <Compile Include="Mono\Funcs.cs" />
   </ItemGroup>
+  <Import Project="Mono.Cecil.settings" />
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
 </Project>
--- a/Mono.Cecil.settings
+++ b/Mono.Cecil.settings
@@ -16,14 +16,14 @@
     <DebugSymbols>true</DebugSymbols>
     <DebugType>full</DebugType>
     <Optimize>false</Optimize>
-    <DefineConstants>DEBUG;TRACE;</DefineConstants>
+    <DefineConstants>$(DefineConstants);DEBUG;TRACE;</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
   </PropertyGroup>
   <PropertyGroup Condition=" $(Configuration.EndsWith('Release')) ">
     <DebugType>pdbonly</DebugType>
     <Optimize>true</Optimize>
-    <DefineConstants>TRACE;</DefineConstants>
+    <DefineConstants>$(DefineConstants);TRACE;</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
     <RunCodeAnalysis>false</RunCodeAnalysis>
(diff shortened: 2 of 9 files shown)


kzu:
@migueldeicaza @alanmcgovern this is one way we wouldn't need to fork cecil anymore just to "own" its assembly name and signing key :)

alanmcgovern:
This is a very neat solution, i like it!

jbevain:
:+1: can you just move the import to the bottom of Mono.Cecil.settings, so it affects the entire project? Thanks!

jbevain:
@kzu you just tweak Mono.Cecil, you don't need Mono.Cecil.Pdb/Mdb?

I suspect not as we have InternalsVisibleTo for those, as that would break.

kzu:
good point :). Will do

kzu:
Ok, moved it to the common .settings file. So it should now be possible to just change the assembly name prefix on all of Cecil by adding an overrides with:

<Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
  <PropertyGroup>
    <AssemblyName>$(AssemblyName.Replace('Mono', 'MyCompany'))</AssemblyName>
    <AssemblyOriginatorKeyFile>$(MSBuildThisFileDirectory)MyCompany.snk</AssemblyOriginatorKeyFile>
  </PropertyGroup>
	<ItemGroup>
		<Compile Include="$(MSBuildThisFileDirectory)Mono.Cecil.AssemblyInfo.cs" />
	</ItemGroup>
</Project>

The custom AssemblyInfo.cs would add the InternalsVisibleTo so that everything compiles just fine.

I just tried it with my own custom key and asm name and it all built beatifully :)

jbevain:
@kzu, nice! If you move the import to Mono.Cecil.settings just before the import of the C# targets, we can make Mono.Cecil.settings import Mono.Cecil.Overrides.settings, and avoid adding another import in the project files.

kzu:
Ah, oki. In fact, if we do that, we don't need the Mono.Cecil.Overrides.settings at all :).

I first tried adding the overrides imports to the settings file at the top, but it didn't override anything, since AssemblyName and stuff was set right after :).

kzu:
Unless you prefer to keep that Overrides.settings file?

jbevain:
@kzu you're right, let's move the import at the bottom and inline Overrides.settings, and merge this!

kzu:
Ok, I think this is the golden one :)

jbevain:
Damn build failure, you jinxed it :)

kzu:
Damn. Moving the settings to the end overwrote the $(DefineConstants) added at the .csproj level. Fixed it and let's see if that fixes the build :)

jbevain:
Thanks!

PR #218, opened 2015-05-27
Minor fix for xBuild compatibility

joj:
Without colons, xBuild will error out on the import (and on the condition). MSBuild likes it, though...

@jbevain this is the change in #4 in mono/cecil.


jbevain:
Thanks!

Issue #209, opened 2015-03-21
Unit test failures on Mono


akoeplinger:
Just tried running the unit tests on Mono 3.12 and got the following failures, are these tests supposed to pass on Mono?

```
Tests run: 234, Errors: 7, Failures: 4, Inconclusive: 0, Time: 15.1602525 seconds
Not run: 0, Invalid: 0, Ignored: 0, Skipped: 0

Errors and Failures:

1) Test Failure : Mono.Cecil.Tests.ImageReadTests.IA64Module
   Error: Invalid PE header Machine valueError count: 1
at Mono.Cecil.Tests.CompilationService.Verify (System.String name) [0x00000] in <filename unknown>:0
at Mono.Cecil.Tests.TestRunner.RoundTrip (System.String location, Mono.Cecil.ReaderParameters reader_parameters, System.String folder) [0x00000] in <filename unknown>:0
at Mono.Cecil.Tests.TestRunner.GetModule () [0x00000] in <filename unknown>:0
at Mono.Cecil.Tests.TestRunner.RunTest () [0x00000] in <filename unknown>:0
at Mono.Cecil.Tests.BaseTestFixture.Run (Mono.Cecil.Tests.TestCase testCase) [0x00000] in <filename unknown>:0
at Mono.Cecil.Tests.BaseTestFixture.TestModule (System.String file, System.Action`1 test, Boolean verify, System.Type symbolReaderProvider, System.Type symbolWriterProvider) [0x00000] in <filename unknown>:0
at Mono.Cecil.Tests.ImageReadTests.IA64Module () [0x00000] in <filename unknown>:0

2) Test Error : Mono.Cecil.Tests.ImageReadTests.MetroAssembly
   Mono.Cecil.AssemblyResolutionException : Failed to resolve assembly: 'System.Runtime, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
  at Mono.Cecil.BaseAssemblyResolver.Resolve (Mono.Cecil.AssemblyNameReference name, Mono.Cecil.ReaderParameters parameters) [0x00000] in <filename unknown>:0 
  at Mono.Cecil.BaseAssemblyResolver.Resolve (Mono.Cecil.AssemblyNameReference name) [0x00000] in <filename unknown>:0 
  at Mono.Cecil.DefaultAssemblyResolver.Resolve (Mono.Cecil.AssemblyNameReference name) [0x00000] in <filename unknown>:0 
[...]

PR #222, opened 2015-06-01
Added travis CI

gluck:
In case you're interested in having mono CI for Cecil, here's the travis conf you can use.
Mono.Cecil.nunit change is compatible with Windows (the old version wasn't working on linux).

Note that there are ~25 UT failures on mono/linux, so enabling travis on Cecil will bring red'ish stickers (rather than green-ish) on PRs.

I also suggest the appveyor config gets commited to the repo in the same way, it's more sane when working with branches and provides a changelog for build configs. You can go to appveyor settings, select "export YAML".

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,8 @@
+language: csharp
+solution: Mono.Cecil.sln
+install:
+  - nuget restore Mono.Cecil.sln
+  - nuget install NUnit.Runners -Version 2.6.4 -OutputDirectory testrunner
+script:
+  - xbuild /p:Configuration=net_4_0_Debug Mono.Cecil.sln
+  - mono ./testrunner/NUnit.Runners.2.6.4/tools/nunit-console.exe Mono.Cecil.nunit
--- a/Mono.Cecil.nunit
+++ b/Mono.Cecil.nunit
@@ -1,9 +1,9 @@
 <NUnitProject>
   <Settings activeconfig="Default" domainUsage="Multiple" />
   <Config name="Default" binpathtype="Auto" runtimeFramework="v4.0.21006">
-    <assembly path="Test\bin\net_4_0_Debug\Mono.Cecil.Tests.dll" />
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -0,0 +1,8 @@
+language: csharp
+solution: Mono.Cecil.sln
+install:
+  - nuget restore Mono.Cecil.sln
```
jbevain: Is this needed? We don't reference nugets in the csprojs.

gluck: Then it's not (but it doesn't fail either, and will be required if one day Cecil decides that some NuGet is worth using).
Travis calls it by default in the install step, so once you override the step (to install NUnit runner), you need to call it explicitely.
TL;DR it can be removed

gluck: (more likely than Cecil using a NuGet package would be to grab NUnit as a NuGet package in UTs, it'll be required then as well)

jbevain:
On the subject of tests on Mono, we recently worked on reducing the failures: https://github.com/jbevain/cecil/issues/209

gluck:
Nice, it seems that there are much less that when I tested last time, you can see the build on my travis setup:
https://travis-ci.org/gluck/cecil/builds/64917435

jbevain:
SimpleInterfaces should now be fixed. The last failing test looks indeed bonkers in terms of IL.

jbevain:
And we should be green on Mono now as well.

gluck:
rebased the PR so that Travis tells us the truth:
https://travis-ci.org/gluck/cecil/builds/64933635

gluck:
Still some failures, though they don't seem to always occur.
The [first run](https://travis-ci.org/gluck/cecil/builds/64917435) on this PR didn't have those.

jbevain:
@gluck they're a result of #225, it's fixed in HEAD.

gluck:
Yet another rebase then, we'll see

gluck:
https://travis-ci.org/gluck/cecil/builds/64949521
:tada:

PR #226, opened 2015-06-01
Compact MetadataSystem dicts after assembly load

gluck:
- not 100% sure about this one, but my assumption was that:
"when using the ImmediateModuleReader, the metadata system
 caches should all be empty (all resolved) after load."
(but I'm not sure that's correct)

- If that holds true, then this change works, and makes sure that
 the (fairly large) dicts gets compacted (by GC'ing them).

### DIFF
--- a/Mono.Cecil/MetadataSystem.cs
+++ b/Mono.Cecil/MetadataSystem.cs
@@ -127,23 +127,23 @@ namespace Mono.Cecil {
 
 		public void Clear ()
 		{
-			if (NestedTypes != null) NestedTypes.Clear ();
-			if (ReverseNestedTypes != null) ReverseNestedTypes.Clear ();
-			if (Interfaces != null) Interfaces.Clear ();
-			if (ClassLayouts != null) ClassLayouts.Clear ();
-			if (FieldLayouts != null) FieldLayouts.Clear ();
-			if (FieldRVAs != null) FieldRVAs.Clear ();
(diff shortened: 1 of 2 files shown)


Mono.Cecil/MetadataSystem.cs
```diff
@@ -127,23 +127,23 @@ static bool TryGetPrimitiveData (TypeReference type, out Row<ElementType, bool>
...
-			if (Overrides != null) Overrides.Clear ();
-			if (CustomAttributes != null) CustomAttributes.Clear ();
-			if (SecurityDeclarations != null) SecurityDeclarations.Clear ();
-			if (Events != null) Events.Clear ();
-			if (Properties != null) Properties.Clear ();
-			if (Semantics != null) Semantics.Clear ();
-			if (PInvokes != null) PInvokes.Clear ();
-			if (GenericParameters != null) GenericParameters.Clear ();
-			if (GenericConstraints != null) GenericConstraints.Clear ();
+			if (NestedTypes != null) NestedTypes = new Dictionary<uint, uint[]> ();
```
jbevain: In that case you should pass a zero capacity to the dictionary.

gluck: That's already the default capacity AFAICT, but I can force it if you prefer.

jbevain:
Fixed in f2097e001c021a59540a889730c04523ebc0d93c. Thanks!

PR #227, opened 2015-06-08
Fix ANR FullName exception

gluck:
With versions w/o major/build no.
The UT is failing w/o the code change (relying on CI to see if other UTs will be impacted)

### DIFF
--- a/Mono.Cecil/AssemblyNameReference.cs
+++ b/Mono.Cecil/AssemblyNameReference.cs
@@ -149,7 +149,11 @@ namespace Mono.Cecil {
 				builder.Append (name);
 				builder.Append (sep);
 				builder.Append ("Version=");
-				builder.Append (version.ToString (fieldCount: 4));
+				builder.Append (version.ToString ());
+				if (version.Build == -1)
+					builder.Append (".0");
+				if (version.Revision == -1)
+					builder.Append (".0");
 				builder.Append (sep);
 				builder.Append ("Culture=");
 				builder.Append (string.IsNullOrEmpty (culture) ? "neutral" : culture);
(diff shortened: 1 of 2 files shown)


jbevain:
That's not correct. The ECMA assembly name doesn't have the same formatting at the .NET Version class. We do need the 4 fields. We need a failing test case and to protect against it.

gluck:
Not sure I what you mean, the included test case is failing with the current code (not because of the no of digits in the output FullName, but because of an exception).
You're saying you'd like this TC to pass, but with 4 digits in the FullName ?

jbevain:
@gluck exactly :)

In retrospect, we should not have used the System.Version type, but that's what we have.

jbevain:
@gluck what I'm really interested in, is how the failing ANR for your bug was created.

gluck:
Here you go !

gluck:
I'm checking the repacked libs (it's one out of them), first guess would be UnionArgParser:
https://www.nuget.org/api/v2/UnionArgParser/0.8.7

Which (strangely) has AssemblyVersion attributed defined twice:

    [assembly: AssemblyVersion("0.8.7.0")]
    [assembly: FSharpInterfaceDataVersion(2, 0, 0)]
    [assembly: Debuggable(DebuggableAttribute.DebuggingModes.IgnoreSymbolStoreSequencePoints)]
    [assembly: AssemblyFileVersion("0.8.7")]
    [assembly: AssemblyVersion("0.8.7")]
    [assembly: TargetFramework(".NETFramework,Version=v4.0", FrameworkDisplayName = ".NET Framework 4")]

jbevain:
That will still be encoded as 0.8.7.0 in the metadata.

gluck:
Ok got it. ILRepack allows to set the version of the generated assembly, generated using `new Version(string)`, likely this one has 3 digits.
What do you recommended ?

jbevain:
Good catch! I've pushed a fix.

Issue #180, opened 2014-09-23
Override method resolution does not find base method with generic parameter


ygoe:
Mono.Cecil.Rocks.MethodDefinitionRocks.GetOriginalBaseMethod() doesn't find the base method of an overridden method if the method has a generic parameter. Here's the offending code:

``` C#
class MyCollection : Collection<string>
{
    // base: InsertItem(int, T)
    protected override void InsertItem(int index, string item) { }
}
```

Since Collection<T>.InsertItem has **T** as its second parameter, and MyCollection.InsertItem has a **string** parameter there, the two parameter lists are not considered equal in Mono.Cecil.MetadataResolver.AreSame(TypeReference, TypeReference) where the etype is compared.

Since one etype says "Var" and the other "String", would it be okay to accept the difference if one is Var? Or would that be too broad?

dg9ngf:
This is really bugging me again now. This very `InsertItem` override method. I have just upgraded from 0.9.5 to 0.9.6 and now it's broken when it wasn't before. I'm using the hack I've described above. Not sure yet why it has worked and doesn't anymore now. Any news on this? It's simply breaking the assemblies I'm working with and I need to find a solution for that. How would this "GetOriginalMethod should check..." look like?

dg9ngf:
I noticed that there is such a thing as explicit method overrides in CIL and it should be usable to find a base method in case this is set. I'm going to do some tests but it seems only possible with C++/CLI not C#. If present, I believe the method.Overrides collection should be checked as well. (I've written some code for that.) But that's probably another issue. And unfortunately, the C# compiler (VS 2010) doesn't set the .override directive in the assembly for overridden methods (of the same name) so I can't use it here. Would be nice and simple. :-)

dg9ngf:
I'm currently in the process of fixing this issue. I can already resolve methods if they are using a single level of GenericParameters. Still trying to get the second level right, like `T1` resolving to `IEnumerator<T2>`. My brain is aching. It's full of generic parameters and generic arguments.

To make this generics resolution truly recursive, I've written a method that only does that resolving of GenericParameters using a list of GenericArguments of some DeclaringType. Unfortunately I can't just replace the GenericParameter I find with its corresponding GenericArgument in the same structure because this affects the entire assembly. There's no Clone method so I can't easily make a copy of the TypeReference.

[...]

PR #229, opened 2015-06-29
#180: Fix: Override method resolution does not find base method with generic parameter

ygoe:
Fixes issue #180


jbevain:
Hi!

Thanks for working on this!

This will need unit tests to ensure we do not regress.

dg9ngf:
Sorry, I won't get to making a test for this. The code works for me though so I'll definitely use it for myself. If anybody wants to see this bug fixed in the official version, they will have to do whatever is necessary theirselves.

jbevain:
Alright then.

PR #230, opened 2015-06-29
Allow setting of TypeSpecification's namespace and name

dg9ngf:
It's not an invalid operation. In fact it's working fine and is required to fix my issue.

### DIFF
--- a/Mono.Cecil/TypeSpecification.cs
+++ b/Mono.Cecil/TypeSpecification.cs
@@ -24,12 +24,12 @@ namespace Mono.Cecil {
 
 		public override string Name {
 			get { return element_type.Name; }
-			set { throw new InvalidOperationException (); }
+			set { element_type.Name = value; }
 		}
 
 		public override string Namespace {
 			get { return element_type.Namespace; }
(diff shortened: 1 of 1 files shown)


jbevain:
Hi!

It is, in fact, an invalid operation :).

If all you have is a TypeSpec you can get its declaring type to alter it, but it doesn't make sense to allow altering a type from its specification.

dg9ngf:
I'm not sure what you mean. I have renamed the referenced type, and it was reflected in the generated assembly. But the other assembly still had a reference to the wrong name and the assembly didn't validate with PEVerify. Changing this fixed the issue. So I believe that my version is correct. Maybe something else in Cecil is broken that it doesn't update the reference where it should.

jbevain:
Sure it works. But from a semantic point of view, it doesn't make any sense :)

Think of a TypeSpec as a view on a Type that adds one annotation. You should not be able to modify the Type through the view.

dg9ngf:
Well, okay. I fixed this for me and thought it may be useful to the project as well. If it's not, I have no problem with that.

PR #238, opened 2015-08-31
fix writting of row/col information in mdb

radekdoulik:
- the arguments to the MarkSequencePoint were passed in a wrong order


jbevain:
Good catch, thanks!

PR #245, opened 2015-11-02
Fix metadata header length calculation

ignas2:
Prior to this fix winmd files were not saved correctly to disk. This is because module's runtime version is longer than 20 bytes.


Test/Mono.Cecil.Tests/BaseTestFixture.cs
```diff
@@ -119,19 +119,21 @@ abstract class TestCase {
 		public readonly bool ReadOnly;
 		public readonly Type SymbolReaderProvider;
 		public readonly Type SymbolWriterProvider;
+		public readonly BaseAssemblyResolver AssemblyResolver;
```
jbevain: Can you use IAssemblyResolver here and in the constructors instead of BaseAssemblyResolver?

jbevain:
That looks great! Please just fix the tidbit and it will be good to merge.

Thanks!

Test/Mono.Cecil.Tests/BaseTestFixture.cs
ignas2: That's what I did first. But IAssemblyResolver doesn't support AddSearchDirectory method which is called by TestRunner.GetModule.

jbevain: That's alright, if the user supplies its own IAssemblyResolver, we can expect that it will be fully configured. The AddSearchDirectory should only be called if a DefaultAssemblyResolver is created.

ignas2:
Done.

jbevain:
Perfect! Thanks a lot!

PR #250, opened 2015-11-29
fixes #248

### LINKED ISSUE #248: Renaming type namespace doesn't update nested type fullname
the following test fails:

```
TypeDefinition ta = new TypeDefinition("ns", "A", TypeAttributes.Class | TypeAttributes.Public);
TypeDefinition tb = new TypeDefinition(null, "B", TypeAttributes.Class | TypeAttributes.Public);
ta.NestedTypes.Add(tb);
string fntb = tb.FullName; //read nested type fullname

ta.Namespace = "nsb";
Assert.IsTrue(tb.FullName.StartsWith("nsb")); //this fails
```

the workaround is to reset nested Namespace to null to force Fullname update.

mwoelk83:


jbevain:
Thanks for this! I fixed it in a slightly different way in c52663e51d1d3ca683c1754fc53119d9707e7437.

PR #251, opened 2015-12-03
Windows Runtime projections

ignas2:
Windows Runtime projections map types from different languages. For example when C++/CX uses Windows::Foundation::Collections::IVector<T> type it appears as System.Collections.Generic.IList<T> to C#.

Could this be resolved with custom MetadataResolver implementation leaving Cecil alone? Kind of, but it's not pretty. For example if you have IClosable.Close method reference it will resolve to IDisposable.Dispose. This confuses a lot of tools which don't expect MethodReference and MethodDefinition to not mach.

Other libraries already support Windows Runtime projections and I tried to emulate one:
https://github.com/dotnet/corefx/blob/master/src/System.Reflection.Metadata/src/System/Reflection/Metadata/MetadataReader.WinMD.cs

I have added just a few public APIs:
- ReaderParameters.ApplyWindowsRuntimeProjections: When set to true metadata is projected to CLR. Otherwise it's loaded as is (which is default).
- ModuleDefinition.MetadataKind: Can have three values: Ecma335 - module is not a winmd file or it has been read with runtime projections disabled (default); WindowsMetadata - native winmd file; ManagedWindowsMetadata - managed winmd file.
- MemberReference.IsWindowsRuntimeProjection - true if Windows Runtime projection has been applied to it; false otherwise.

I have also tried to make so that there would be no performance impact for assemblies that don't care about projections. We could #ifdef all this code if needed.

Tests are missing but will be added.

It would be nice to expose underlying type/method/etc for projected type/method/etc.

Projected metadata can't be modified at the moment.

Let me know what you think. I'm happy to change things around (or everything if needed) as I really want to see this feature in Cecil. I've tried using this build with Mono Linker and it works fine with winmd files with very few tweaks.

### DIFF
--- a/Mono.Cecil.csproj
+++ b/Mono.Cecil.csproj
@@ -55,8 +55,10 @@
     <Compile Include="Mono.Cecil\AssemblyWriter.cs" />
     <Compile Include="Mono.Cecil\BaseAssemblyResolver.cs" />
     <Compile Include="Mono.Cecil\CallSite.cs" />
+    <Compile Include="Mono.Cecil\Treatments.cs" />
     <Compile Include="Mono.Cecil\TypeParser.cs" />
     <Compile Include="Mono.Cecil\Import.cs" />
+    <Compile Include="Mono.Cecil\WindowsRuntimeProjections.cs" />
     <Compile Include="Mono.Collections.Generic\Collection.cs" />
     <Compile Include="Mono.Cecil\ExportedType.cs" />
     <Compile Include="Mono.Cecil\SecurityDeclaration.cs" />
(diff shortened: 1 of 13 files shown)


jbevain:
Wow, again this looks like great work, I'll review ASAP.

Sorry for the lag processing your contributions, I've just been relocating halfway through the globe, so I have very little time for myself.

ignas2:
No worries. Do you have ETA when you'll be able to review this?

ignas2:
Ping? I'll be relentless on this one :)

jbevain:
Can you resubmit a PR targeting jbevain:winrt-projections? Thanks!

PR #255, opened 2016-01-19
This file doesn't use any markdown

ghost:
... but the fact that it has the copyright holders' names on
consecutive lines means that interpreting it as markdown will
concatenate them onto the same line.


jbevain:
Thanks :)

PR #258, opened 2016-02-22
add equality comparing methods for TypeReference

Licshee:
when using ToDefinition() from the test projects, the definition objects are are not same as their "nature" equivalents, this makes testing code involves TypeDefinition relations hard

so i made these methods as a workaround


Licshee:
i was intended to make a separate NuGet package, but the Cecil.nupkg does not support PCL
could also add Equals & operator implementations, but not sure about your ideal about designing the API
so i finally decided to make changes to Rocks

Licshee:
broken

PR #261, opened 2016-02-25
Delete the unused property

soukouki:


jbevain:
Hi,

Thanks for the PR. It's not clear to me what your intent is.

Why would you rename this property? Removing the properties is a breaking change.

Jb

PR #263, opened 2016-03-07
Support WinRT module import

gluck:
No experience in UWP, but somebody hit that one in ILRepack:
https://github.com/gluck/il-repack/issues/150


jbevain:
Looks good, thanks!

PR #249, opened 2015-11-18
Custom attribute support for interface implementations and type references

ignas2:

PR #268, opened 2016-04-21
Support reading and writing custom attributes on InterfaceImpl

jbevain:
WinRT uses custom attributes in winmd files that are declared for InterfaceImpl rows.
This is a breaking API change that we could not represent before.
There's other places that we do not support reading and writing custom attributes on,
like generic parameter constraints. Let's fix it when we need it.

### DIFF
--- a/Mono.Cecil/AssemblyReader.cs
+++ b/Mono.Cecil/AssemblyReader.cs
@@ -1123,25 +1123,29 @@ namespace Mono.Cecil {
 		public bool HasInterfaces (TypeDefinition type)
 		{
 			InitializeInterfaces ();
-			MetadataToken [] mapping;
+			Row<uint, MetadataToken> [] mapping;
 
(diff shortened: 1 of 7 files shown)


jbevain:
This approach is competing with #249.

jbevain:
The cons:
 * API breaking change,
* One indirection for the common path, getting the type of the interfaces,
* Only supports InterfaceImpl.

The pros:
* More inline with the DOM approach of Cecil.
* Easier to maintain.

TautvydasZilys:
I like this way more. I remember the discussion we had in the office about how to expose interface impls, and we made a very similar implementation to this:

https://github.com/ignas2/cecil/commit/1f46149000eba8affd1fe61ab76c9eb2c96160ed

However, since it broke API, we dismissed it quite quickly. If you think breaking API is ok, I'd suggest we stick with your changes.

PR #271, opened 2016-05-12
Lower memory usage

jbevain:
The way Cecil works right now is that it reads the entire assembly into memory and reads its object model from that. If you have a large assembly with a lot of IL and especially large resource files, this has a tendency of quickly consuming lots of memory.

This fix introduces an important API breaking change and concept.

ModuleDefinition is now IDisposable. Everything that holds onto ModuleDefinitions becomes IDisposable as well: AssemblyDefinition, DefaultAssemblyResolver.

**The default implementation preserves the old behavior and reads the assembly in memory, thus you do not need to explicitely call dispose on your modules**. If you want to lower the memory usage of your tools and are okay with keeping handles open on the files, pass:

```csharp
new ReaderParameters {
    InMemory = false,
}
```

When reading a module.


SimonCropp:
very timely https://github.com/Fody/Costura/issues/154

tmat:
@jbevain You might find these helpers useful: https://github.com/dotnet/corefx/tree/master/src/System.Reflection.Metadata/src/System/Reflection/Internal/MemoryBlocks. We focused on optimizing memory allocations quite a bit for the metadata reader used by Roslyn.

Issue #90, opened 2012-02-29
pdb roundtrip loses information


jbevain:

jbevain:
There's a few issues at hand here. Round-tripping pdb files makes it so that we're losing some of the «custom metadata» of the pdb.

The plan to fix that I'm working on is as follows:

https://groups.google.com/forum/#!topic/mono-cecil/VSGWjgwDC7Y

I'm slowly putting more and more work into Cecil after an intense period of professional activity.

If a company using Cecil wants to speed things up, they can sponsor the work.

user1568891:
Any progress on this? I've always wondered why the watch and immediate window never worked properly on my project that uses Fody. Finally found the answer on [StackOverflow] (http://stackoverflow.com/a/9494198/1568891) that sent me here.

drauch:
Had to learn it the hard way also: https://stackoverflow.com/questions/34267811/vs2015-linq-in-debugger-watch-window

Would be nice if there is any follow-up on this :+1:

xen2:
FYI, started to work on it again yesterday.
Promising results so far, hopefully a patch within a few days.
@jbevain, can you confirm that you didn't work on it yet so as to not overlap? (according to https://groups.google.com/d/topic/mono-cecil/VSGWjgwDC7Y/discussion)

user1568891:
Thanks!

drauch:
@xen2: Thank you for your time! :+1: Please let us know if we can help, e.g. testing, ...

xen2:
@drauch If interested, you can cherry-pick 214b1a69a2d87d8cf64edca0d214f7816395de24 (early testing version, still various things to improve/cleanup/fix).

xen2:
Quick update, 455f850 should be very close to final.

Only thing I wasn't really sure of, is why I need to sometime use offset and sometimes offset + 1 for my iterator scope end, but it seems to work as is.

The cool thing is I didn't do only the async locals but I also added support for many other PDB info:
- Using clauses (before `typeof(Console)` was broken in VS Watch window even when `using System` was in your file)
- Stepping over await code works (it will wait for the await call to be finished and go to next line instead of stepping out)

davidroth:
@xen2 Cool! Thx for the fix! 
I have just quick tested it with Fody and works very well so far!
Very nice to have full debugging support back :)

@SimonCropp Would be cool if you could upgrade fody with this patched lib once it is tested.

SimonCropp:
[...]

PR #273, opened 2016-05-17
PDB roundtrip

xen2:
Following #90, opening a pull request to discuss/merge the changes.

On our side, the only problem we still have is, if you have a local var with same name defined twice in your function, it sometimes show the wrong one in the Locals/Watch window.


xen2:
Update: I added PDB scope info
This allows locals to be scoped (and have proper mouse override behavior in case you have multiple variable with same name in different scopes)

Note that for now, some info is still stored at method level but should be at scope level (i.e. using clause per scope or yield iterator in a scope (?), this is not really used much).
That can be done in a separate PR later though.

davidroth:
@xen2  Whats the status of this PR? Are you still working on this? (There are some merge conflicts atm)
@jbevain Any plans to merge this?

xen2:
Closing in favor of #351

PR #279, opened 2016-07-13
Validate PDB magic and ImageDebugDirectory type and version

AustinWise:
Presently PdbReader does not validate that it is actually opening a traditional PDB. If a portable PDB is opened, the Portable PDB file is blindly assumed to be a regular PDB and parsing eventually fails when a huge array is attempted to be allocated:
```
System.OutOfMemoryException: Array dimensions exceeded supported range.
   at Microsoft.Cci.Pdb.MsfDirectory..ctor(PdbReader reader, PdbFileHeader head, BitAccess bits)
   at Microsoft.Cci.Pdb.PdbFile.LoadFunctions(Stream read, Dictionary`2& tokenToSourceMapping, String& sourceServerData, Int32& age, Guid& guid)
   at Mono.Cecil.Pdb.PdbReader.PopulateFunctions()
   at Mono.Cecil.Pdb.PdbReader.ProcessDebugHeader(ImageDebugDirectory directory, Byte[] header)
   at Mono.Cecil.ModuleDefinition.ProcessDebugHeader()
   at ICSharpCode.ILSpy.LoadedAssembly.LoadSymbols(ModuleDefinition module)
```
This pull request validates both the ImageDebugDirectory in the PE file and the magic at the beginning of a PDB file. This change is motivated by icsharpcode/ILSpy#723, which expects `InvalidOperationException` when the file is invalid, not `OutOfMemoryException`.

### DIFF
--- a/symbols/pdb/Microsoft.Cci.Pdb/PdbFileHeader.cs
+++ b/symbols/pdb/Microsoft.Cci.Pdb/PdbFileHeader.cs
@@ -24,6 +24,8 @@ namespace Microsoft.Cci.Pdb {
     //  this.pageSize = pageSize;
     //}
 
+    const string MAGIC = "Microsoft C/C++ MSF 7.00";
+
     internal PdbFileHeader(Stream reader, BitAccess bits) {
       bits.MinCapacity(56);
       reader.Seek(0, SeekOrigin.Begin);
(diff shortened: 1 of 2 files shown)


jbevain:
Thanks Austin!

PR #293, opened 2016-09-22
Use debug symbol files based on files availability instead of hardcoded values

marek-safar:
Enables implicit (p)pdb support on mono


jbevain:
Closed in favor of #322. Thanks dude!

PR #301, opened 2016-10-23
Changed PdbReaderProvider.IsPortablePdb so that it uses FileShare.Read

vanderkleij:
I've updated [Smocks](https://github.com/vanderkleij/Smocks) to use Cecil 0.10.0-beta1-v2. Many of its unit tests then broke because Cecil's `PdbReaderProvider` was trying to open PDB files that were already in use by another process (probably Visual Studio) with exclusive access (FileShare.None).

Since we're only reading from the PDB, I don't think exclusive access is required and we can just allow others to read from the PDB simultaneously (FileShare.Read).


jbevain:
That makes sense! Thanks for the PR, this will go into 0.10-beta2!

Issue #309, opened 2016-12-14
Wrong portable pdb minor version


akoeplinger:
Cecil uses the following versions for portable pdb (from [PortablePdb.cs#L207](https://github.com/jbevain/cecil/blob/c0181b8b3c091efd6bc5f4d2e5888d0f35b4fef7/Mono.Cecil.Cil/PortablePdb.cs#L207)

```
directory = new ImageDebugDirectory () {
	MajorVersion = 256,
	MinorVersion = 20577,
	Type = 2,
};
```

The minor version seems to be wrong, it should be 20557 (0x504d) according to the spec: https://github.com/dotnet/corefx/blob/fd15ab1de03a4431e0ead9ff10e1bf53e2ae4bb7/src/System.Reflection.Metadata/specs/PE-COFF.md#codeview-debug-directory-entry-type-2

@jbevain Looks like a simple typo. I noticed this because a cecil-processed assembly wouldn't load the .pdb on Mono as the minor version doesn't match.

PR #310, opened 2016-12-14
Fix ImageDebugDirectory.MinorVersion in portable pdb writer

akoeplinger:
The minor version is wrong, it should be 20557 (0x504d) according to the spec: https://github.com/dotnet/corefx/blob/fd15ab1de03a4431e0ead9ff10e1bf53e2ae4bb7/src/System.Reflection.Metadata/specs/PE-COFF.md#codeview-debug-directory-entry-type-2

Fixes https://github.com/jbevain/cecil/issues/309


jbevain:
Thanks!

PR #313, opened 2017-01-08
[ppdb] Fix the encoding/decoding of empty parts in document names, th…

vargaz:
…ey should be encoded as index 0. Also, empty parts should be retained during writing, since /foo/bar is encoded as <empty><foo><bar>.


vargaz:
This problem currenty affects the mono build, where stacktraces would show up as:
at System.Threading.Thread.Sleep (int) [0x00019] in
Users/vargaz/git/mono/mcs/class/referencesource/mscorlib/system/threading/thread.cs:741

(note the missing / from the path).

jbevain:
Thanks for the PR, this is good to go, could you just add a small test to:

https://github.com/jbevain/cecil/blob/master/Test/Mono.Cecil.Tests/PortablePdbTests.cs

Thanks!

vargaz:
Those tests look windows only.

jbevain:
They are not windows only (for some the verifier is only ran on windows), you can just upload a small binary to Test/Resources/assemblies with the symbol file, and add a test to PortablePdbTest ensuring we round-trip the unix path in ppdb nicely :)

vargaz:
Done.

jbevain:
Thanks!

PR #320, opened 2017-01-23
Fix BaseAssemblyResolver.SearchDirectory.

erozenfeld:
Fix BaseAssemblyResolver.SearchDirectory to try both .exe and .dll when searching for an assembly.
This fixes a scenario when there is a native Foo.exe and a managed Foo.dll in the search directories.

### DIFF
--- a/Mono.Cecil/BaseAssemblyResolver.cs
+++ b/Mono.Cecil/BaseAssemblyResolver.cs
@@ -158,8 +158,13 @@ namespace Mono.Cecil {
 			foreach (var directory in directories) {
 				foreach (var extension in extensions) {
 					string file = Path.Combine (directory, name.Name + extension);
-					if (File.Exists (file))
+					if (!File.Exists (file))
+						continue;
+					try {
 						return GetAssembly (file, parameters);
+					} catch (System.BadImageFormatException) {
+						continue;
+					}
 				}
 			}
 


Mono.Cecil/BaseAssemblyResolver.cs
```diff
@@ -158,8 +158,14 @@ AssemblyDefinition SearchDirectory (AssemblyNameReference name, IEnumerable<stri
 			foreach (var directory in directories) {
 				foreach (var extension in extensions) {
 					string file = Path.Combine (directory, name.Name + extension);
-					if (File.Exists (file))
-						return GetAssembly (file, parameters);
+					if (File.Exists (file)) {
```
jbevain: Could you invert the condition and continue if there's no file to reduce indentation?

Mono.Cecil/BaseAssemblyResolver.cs
```diff
@@ -158,8 +158,14 @@ AssemblyDefinition SearchDirectory (AssemblyNameReference name, IEnumerable<stri
...
+						try {
+							return GetAssembly (file, parameters);
+						}
+						catch (System.BadImageFormatException) {
```
jbevain: In the Mono coding conventions, the catch goes on the same line as the closing } for the try, like this:

```csharp
} catch {
```

jbevain:
Hey @erozenfeld,

Thanks for the PR! This indeed makes a lot of sense. Could you please address the couple of tidbits comments I added and keep the PR with just 1 commit by squashing your fix, and this is good to go.

Thanks!

erozenfeld:
@jbevain Thank you for the review. I addressed your comments.

jbevain:
Thanks!

PR #322, opened 2017-01-28
Add best effort reading and writing of debug symbols

jbevain:
Before the portable pdb format existed, we had one platform associated with one type of debug symbols. The «best effort» reading of Cecil, expressed through:

```csharp
new ReaderParameters { ReadSymbols = true }
```

Was trying to load a .mdb if the code was running on Mono, or trying to load a .pdb if the code was running on .NET. Not only this doesn't work anymore because of the portable pdb, but it was annoying to use if you were writing a general purpose tool and didn't know what kind of assembly and symbol you could receive.

This PR adds a new DefaultSymbolReaderProvider that only supports file based symbols. The DefaultSymbolReaderProvider will read the first symbol that is along with the .dll.

This PR adds a new DefaultSymbolWriterProvider that will write the kind of format that was read.

Doing

```csharp
new ReaderParameters { ReadSymbols = true }
```

Is now equivalent to:

```csharp
new ReaderParameters { SymbolReaderProvider = new DefaultSymbolReaderProvider () }
```

Same for WriteSymbol/DefaultSymbolWriterProvider.


marek-safar:
Shouldn't `new ReaderParameters { SymbolProvider = new DefaultSymbolReaderProvider () }`
be `new ReaderParameters { SymbolReaderProvider = new DefaultSymbolReaderProvider () }` instead

jbevain:
@marek-safar you're technically correct, the best kind of correct!

Mono.Cecil.Cil/Symbols.cs
```diff
@@ -630,16 +630,52 @@ public interface ISymbolReaderProvider {
...
+		public DefaultSymbolReaderProvider (bool throwIfNoSymbol = true)
+		{
+			throw_if_no_symbol = throwIfNoSymbol;
+		}
+
+		public ISymbolReader GetSymbolReader (ModuleDefinition module, string fileName)
+		{
+			var pdb_file_name = Mixin.GetPdbFileName (fileName);
+			if (File.Exists (pdb_file_name))
+				return SymbolProvider.GetReaderProvider (SymbolKind.Pdb).GetSymbolReader (module, fileName);
```
marek-safar: This is wrong because it always tries to create `Mono.Cecil.Pdb.PdbReaderProvider` which means native reader implementation has to be available to read portable pdb.
Secondly, GetReaderProvider can returns null as valid value which should be handled probably by falling to Mdb and skipping the throw

jbevain: Yep, I agree with your review. Thanks!

jbevain:
@marek-safar we should be in a pretty good place, could you give this a try?

marek-safar:
@jbevain this version works for me

jbevain:
That will be in 0.10 beta 3.

PR #324, opened 2017-01-31
Fixed possible NullReferenceException

mmayr-at:
I have some assemblies without a `ResourceTable` entry, which produce a `System.NullReferenceException` ~~and no types are listed in the ILSpy TreeView~~. I've added a null check to fix this.

### DIFF
--- a/Mono.Cecil/AssemblyReader.cs
+++ b/Mono.Cecil/AssemblyReader.cs
@@ -656,6 +656,9 @@ namespace Mono.Cecil {
 		{
 			var rva = image.Resources.VirtualAddress;
 			var section = image.GetSectionAtVirtualAddress (rva);
+			if (section == null)
+				return null;
+
 			var position = (rva - section.VirtualAddress) + offset;
 			var buffer = section.Data;
 


jbevain:
Hey, thanks for the fix! Could you make this PR against master instead of 0.9.6?

mmayr-at:
There is already a proper null check in `Image.GetReaderAt` in master. I made this PR against this branch, cause ILSpy is using this version.

jbevain:
Do you guys need a new release of the 0.9.6 with this patch? Any roadblock upgrading to 0.10?

mmayr-at:
A new version may not be necessary, because the ILSpy repo includes Cecil as git submodule. I don't know, if the ILSpy devs have any upgrade plans, but I could try to make a PR for this. Feel free to close this PR

jbevain:
Looking forward to having ILSpy switch to 0.10.

@mmayr-at please use the list of breaking changes in [Cecil 0.10](http://cecil.pe/post/149243207656/mono-cecil-010-beta-1) when porting the code base. Don't hesitate to get in touch if you have any question or issue.

Issue #314, opened 2017-01-09
Optimize StringHeap size


erozenfeld:
String suffixes can be used to optimize the size of StringHeap. For example if `foobar` and `bar` are needed, only `foobar` has to be emitted, its suffix can be used for `bar`. Here is how this is implemented in Roslyn: https://github.com/dotnet/roslyn/blob/614299ff83da9959fa07131c6d0ffbc58873b6ae/src/Compilers/Core/Portable/System/Reflection/Metadata/Ecma335/MetadataBuilder.Heaps.cs#L287

jbevain:
@erozenfeld cool, are you working on adding support for this to Cecil?

erozenfeld:
@jbevain Not yet but I can probably get to this later this week.

jbevain:
Sounds great, looking forward to reviewing the PR :)

erozenfeld:
This was implemented in #326.

PR #326, opened 2017-02-10
Optimize StringHeap size.

erozenfeld:
This change reduces the size of StringHeap by re-using string suffixes wherever possible.
Before this change, if metadata strings `foo` and `foobar` were needed, both would be emitted.
With this change, only `foobar` will be emitted and its suffix will be used for `bar`.

Most of the implementation is taken from Roslyn: https://github.com/dotnet/roslyn/blob/614299ff83da9959fa07131c6d0ffbc58873b6ae/src/Compilers/Core/Portable/System/Reflection/Metadata/Ecma335/MetadataBuilder.Heaps.cs#L287

StringHeapBuffer.GetStringIndex doesn't eagerly write bytes to the buffer; instead, it just assigns each string
an index and keeps the map from strings to indexes. After AssemblyWriter.BuildModule completes and all strings have been
collected, a new method WriteStrings implements the optimization that reuses string suffixes. A map of string indexes to buffer
offsets is returned. That map is used in Metadata.WriteString to emit correct string offsets.

### DIFF
--- a/Mono.Cecil.Cil/PortablePdb.cs
+++ b/Mono.Cecil.Cil/PortablePdb.cs
@@ -249,6 +249,7 @@ namespace Mono.Cecil.Cil {
 				return;
 
 			WritePdbHeap ();
+
 			WriteTableHeap ();
 
 			writer.BuildMetadataTextMap ();
@@ -291,6 +292,10 @@ namespace Mono.Cecil.Cil {
 
 		void WriteTableHeap ()
 		{
+			if (module_metadata.string_heap != pdb_metadata.string_heap) {
+				pdb_metadata.table_heap.string_offsets = pdb_metadata.string_heap.WriteStrings ();
+			}
+
 			pdb_metadata.table_heap.WriteTableHeap ();
 		}
 	}
--- a/Mono.Cecil.Metadata/Buffers.cs
+++ b/Mono.Cecil.Metadata/Buffers.cs
@@ -34,6 +34,8 @@ namespace Mono.Cecil.Metadata {
 		readonly int [] coded_index_sizes = new int [Mixin.CodedIndexCount];
 		readonly Func<Table, int> counter;
 
+		internal uint [] string_offsets;
+
 		public override bool IsEmpty {
 			get { return false; }
 		}
@@ -81,7 +83,7 @@ namespace Mono.Cecil.Metadata {
 
 		public void WriteString (uint @string)
 		{
-			WriteBySize (@string, large_string);
+			WriteBySize (string_offsets [@string], large_string);
 		}
 
 		public void WriteBlob (uint blob)
@@ -299,7 +301,7 @@ namespace Mono.Cecil.Metadata {
 
 	class StringHeapBuffer : HeapBuffer {
 
-		readonly Dictionary<string, uint> strings = new Dictionary<string, uint> (StringComparer.Ordinal);
+		protected Dictionary<string, uint> strings = new Dictionary<string, uint> (StringComparer.Ordinal);
 
 		public sealed override bool IsEmpty {
 			get { return length <= 1; }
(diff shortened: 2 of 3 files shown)


erozenfeld:
Fixes #314.

erozenfeld:
@jbevain @spouliot @marek-safar PTAL

erozenfeld:
Portable pdb tests are failing, so this needs more work. I'll investigate.

erozenfeld:
I pushed a fix, tests are passing now. @jbevain @spouliot @marek-safar PTAL

Mono.Cecil.Cil/PortablePdb.cs
```diff
@@ -249,6 +249,12 @@ public void Dispose ()
 				return;
 
 			WritePdbHeap ();
+
+			if (module_metadata.string_heap != pdb_metadata.string_heap) {
```
jbevain: Isn't this condition always true?
Also please move this to the local WriteTableHeap method.

Mono.Cecil.Metadata/Buffers.cs
```diff
@@ -34,6 +34,8 @@ sealed class TableHeapBuffer : HeapBuffer {
 		readonly int [] coded_index_sizes = new int [Mixin.CodedIndexCount];
 		readonly Func<Table, int> counter;
 
+		internal uint [] stringOffsets;
```
jbevain: Should be string_offsets

Mono.Cecil.Metadata/Buffers.cs
```diff
@@ -311,23 +313,81 @@ public StringHeapBuffer ()
...
 				return index;
 
-			index = (uint) base.position;
-			WriteString (@string);
+			index = (uint) strings.Count + 1;
 			strings.Add (@string, index);
 			return index;
 		}
 
+		public uint[] WriteStrings()
```
jbevain: `public uint [] WriteStrings ()`

Mono.Cecil.Metadata/Buffers.cs
```diff
@@ -311,23 +313,81 @@ public StringHeapBuffer ()
...
+		{
+			var sorted = new List<KeyValuePair<string, uint>> (strings);
+			sorted.Sort (new SuffixSort ());
+			strings = null;
+
+			// Add 1 for empty string whose index and offset are both 0
+			uint [] stringOffsets = new uint [sorted.Count + 1];
```
jbevain: `var string_offsets`

Mono.Cecil.Metadata/Buffers.cs
```diff
@@ -311,23 +313,81 @@ public StringHeapBuffer ()
...
+			stringOffsets [0] = 0;
+
+			// Find strings that can be folded
+			string prev = string.Empty;
```
jbevain: `var previous`

Mono.Cecil.Metadata/Buffers.cs
```diff
@@ -311,23 +313,81 @@ public StringHeapBuffer ()
...
+
 		protected virtual void WriteString (string @string)
 		{
 			WriteBytes (Encoding.UTF8.GetBytes (@string));
 			WriteByte (0);
 		}
+
+		// Sorts strings such that a string is followed immediately by all strings
+		// that are a suffix of it.  
+		private class SuffixSort : IComparer<KeyValuePair<string, uint>>
```
jbevain: Opening { goes on the same line for classes.

Mono.Cecil.Metadata/Buffers.cs
```diff
@@ -311,23 +313,81 @@ public StringHeapBuffer ()
...
+		{
+			public int Compare(KeyValuePair<string, uint> xPair, KeyValuePair<string, uint> yPair)
```
jbevain: `public int Compare (`

Mono.Cecil.Metadata/Buffers.cs
```diff
@@ -311,23 +313,81 @@ public StringHeapBuffer ()
...
+			{
+				string x = xPair.Key;
```
jbevain: var

Mono.Cecil.Metadata/Buffers.cs
```diff
@@ -365,6 +425,18 @@ void WriteBlob (ByteBuffer blob)
 
 	sealed class UserStringHeapBuffer : StringHeapBuffer {
 
+		public override uint GetStringIndex(string @string)
```
jbevain: `GetStringIndex (string`

jbevain:
Thanks for your PR, great job! I've added a bunch of style comments that need fixing-up.

Mono.Cecil.Cil/PortablePdb.cs
erozenfeld: Can't `module_metadata` and `pdb_metadata` be the same? SetModuleMetadata has a check for that so I assumed it was possible. I moved the check to WriteTableHeap.

erozenfeld:
@jbevain I addressed all of your comments. PTAL.

Mono.Cecil.Metadata/Buffers.cs
```diff
@@ -311,23 +313,79 @@ public StringHeapBuffer ()
...
 
+		public uint [] WriteStrings()
+		{
+			var sorted = new List<KeyValuePair<string, uint>> (strings);
+			sorted.Sort (new SuffixSort ());
+			strings = null;
+
+			// Add 1 for empty string whose index and offset are both 0
+			var string_offsets = new uint [sorted.Count + 1];
+			string_offsets [0] = 0;
```
jbevain: That's a no-op.

erozenfeld: Sure, but I think it's better to leave it to make it clear that having 0 there is the intent.

Mono.Cecil.Cil/PortablePdb.cs
jbevain: SetModuleMetadata is called for the embedded and external scenarios. PortablePdb.Dispose() returns early if embedded, so it should be unnecessary.

erozenfeld: My preference is to be defensive and keep the check or at least have an assert but I can remove it if you insist.

jbevain: We're good, the changes are approved :)

jbevain:
Thanks for the PR, this will go into 0.10-beta 3.

PR #327, opened 2017-02-14
Throw detailed exception when debug symbols provider is not found

marek-safar:
instead of NullReferenceException in Mono.Cecil.Cil.SymbolProvider.GetReaderProvider


jbevain:
That is indeed much better. Thanks!

PR #328, opened 2017-02-15
enable build using new cli tools

sbomer:
@erozenfeld @jbevain 

This change adds new .csproj files designed for consumption using the
new dotnet cli tools. This enables one to build cecil using the commands
```
dotnet restore Mono.Cecil.new.csproj
dotnet build Mono.Cecil.new.csproj
```

The new .csproj file defines the build constant NET_CORE when cecil is
being built against netstandard (the new file also supports building
for the desktop framework).

Because the new .csproj is used to specify some assembly metadata that
was previously in .cs files, the redundant information has been put
under an #ifdef in the .cs files.

Finally, one line has been changed to use GetTypeInfo, a reflection
API in netstandard1.3 but not in net45.

### DIFF
--- a/Mono.Cecil.Cil/Symbols.cs
+++ b/Mono.Cecil.Cil/Symbols.cs
@@ -12,6 +12,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Runtime.InteropServices;
+using System.Reflection;
 using SR = System.Reflection;
 
 using Mono.Collections.Generic;
@@ -739,7 +740,11 @@ namespace Mono.Cecil.Cil {
 		static string GetSymbolTypeName (SymbolKind kind, string name)
 		{
 			var ns = GetSymbolNamespace (kind);
+#if !NET_CORE
 			return typeof (SymbolProvider).Assembly.GetName ().Name + "." + ns + "." + kind + name;
+#else
+			return typeof (SymbolProvider).GetTypeInfo().Assembly.GetName ().Name + "." + ns + "." + kind + name;
+#endif
 		}
 
 		static string GetSymbolNamespace (SymbolKind kind)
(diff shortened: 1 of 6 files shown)


jbevain:
I've always resisted creating multiple versions of csproj. The core folder is a temporary hack. This needs to be integrated within our existing system.

sbomer:
Thanks @jbevain.

Would it be acceptable to update the old .csproj to use the new format, with `<Project Sdk="Microsoft.NET.Sdk">` at the top, and implicit references to the sdk and netstandard packages? The downside is that everyone consuming the new .csproj would have to update to an RC version of the cli.

Perhaps another option would be to leave the old-style csproj in cecil, and find a way to depend on it from projects using the new cli tools. Do you know if this is possible?

Thoughts?

jbevain:
That will probably require tinkering around, but what I would like to achieve is participating in our existing MSBuild based build definitions:

See: https://github.com/jbevain/cecil/blob/master/Mono.Cecil.props

Ideally we would simply need to have netstandard_Debug and netstandard_Release configurations that define only the required properties.

sbomer:
@jbevain, I spent some time investigating this, and I don't see a way to target netstandard from the existing msbuild definitions. The issue is that targeting netstandard is currently only supported with the new dotnet cli tools, and these tools expect a new-style .csproj.

I see two options that would enable consumption of Mono.Cecil from .NET core apps using the new tooling:

- Make a single new-style .csproj for Mono.Cecil that supports all of the old configurations currently in Mono.Cecil.props in addition to the netstandard config. This has the downside of only being usable with the new cli tools.
- Make a temporary new-style .csproj just for netstandard. After the new cli tools are released, we could make this the default and integrate the existing configurations into a combined new-style .csproj.

Maybe another option is for us to depend on our own fork of Mono.Cecil until the official repo supports netstandard. @erozenfeld, what do you think?

jbevain:
I have a branch here where I can do:

https://github.com/jbevain/cecil/tree/netstandard

dotnet restore Mono.Cecil.csproj /p:Configuration=netstandard_Debug
dotnet build Mono.Cecil.csproj  /p:Configuration=netstandard_Debug

jbevain:
Closed in favor of #330.

sbomer:
Thanks for looking into this! :)
I guess the trouble I was running into had nothing to do with netstandard.

PR #330, opened 2017-02-16
Support netstandard as part of the standard msbuild tooling

jbevain:
We currently build the netstandard assembly through the temporary project. json hack.
This allows us to reuse the existing multi-target msbuild infrastructure to target netstandard as well.


sbomer:
Hey @jbevain, thanks again for looking into this.

When I build your netstandard branch on linux in the netstandard_Debug configuration, I'm not able to reference Mono.Cecil.dll. I get the following message when I try to build a project with a P2P reference to Mono.Cecil.csproj:

`CSC : error CS0009: Metadata file '/home/sven/src/dotnet/linker/cecil/bin/netstandard_Debug/netstandard1.3/Mono.Cecil.dll' could not be opened -- PE image doesn't contain managed metadata.`

Looking at the dll with ildasm shows the same thing:
```
~/src/dotnet/coreclr/bin/Product/Linux.x64.Debug/ildasm ../cecil/bin/netstandard_Debug/netstandard1.3/Mono.Cecil.dll

//  Microsoft (R) .NET Framework IL Disassembler.  Version 4.5.22220.0

error : File '../cecil/bin/netstandard_Debug/netstandard1.3/Mono.Cecil.dll' not found or not a PE file
```

This works when I build the same project on windows.
Do you know what's going on?

jbevain:
If it works on Windows but not on Linux then it sounds like a .NET Core Linux bug.

sbomer:
After doing a git clean, I had to make one more change to NetStandard.props to build and sign with mono.snk on linux:

`<PublicSign Condition=" '$(OS)' != 'Windows_NT' ">true</PublicSign>`

With this change I can inspect the output assembly using ildasm. :)

jbevain:
Makes sense. Roslyn can only public sign on !Windows. I'll add that.

jbevain:
@sbomer so, does this work for you?

sbomer:
This works to build my project with a P2P reference to cecil. I'm currently investigating an issue with rocks - I think to build rocks against netstandard, we should put SecurityDeclarationRocks.cs under an #if !NET_CORE. With that change I can build rocks against netstandard as well, but I'm still running into an issue where the build doesn't properly set DefineConstants to include INSIDE_ROCKS for some reason.

jbevain:
Right, importing NetStandard.props will overwrite the existing defines for some reasons, which is why I got them directly inside this file.

What do you need Rocks for? The linker only uses Mono.Cecil.dll, and so far we haven't shipped a netstandard Mono.Cecil.Rocks.dll.

sbomer:
I'm depending on rocks for the linker xunit tests. If we don't want to support netstandard on rocks, I can probably remove this dependency.

jbevain:
Yeah I think you're better off with getting the files in Rocks you need and using that in your project.

sbomer:
Ok, I'll do that. In that case this change LGTM.
Thanks!

sbomer:
@jbevain, do I need to do anything to get this into mono/cecil?

jbevain:
You probably need to summon @marek-safar.

marek-safar:
mono/cecil has been updated

sbomer:
Thank you!

PR #333, opened 2017-02-21
Preserve PE timestamps.

rolfbjarne:
Read PE timestamps from PE files, and write the same value back.

This makes Cecil's output deterministic when used with C# or VB's
/deterministic compiler option.


jbevain:
This looks good to me, thank you!

PR #335, opened 2017-02-23
Support ReadSymbols() on .NET Core

sbomer:
It looks like [322](https://github.com/jbevain/cecil/pull/322) changed ReadSymbols() to use DefaultSymbolReaderProvider, so we should be able to use ReadSymbols() on .NET Core now.


jbevain:
Thanks!

sbomer:
@marek-safar, could you get this into mono/cecil as well?
Thanks!

marek-safar:
@sbomer done

PR #336, opened 2017-02-24
Changes made after analysis by LOC8, a Lack of Cohesion tool that is …

zebmason:
…under development. i.e. I am training my tool on other peoples code and coding styles.

N.B. Code has only been checked to see that it compiles.

Hopefully this tidies up a little code by deleting it. If the code was useful then maybe wrap with #if DEBUG.

HTH

Zeb


jbevain:
Thanks for this contribution! I won't merge the PR as is but look into what it makes sense to pick.

PR #340, opened 2017-02-28
Make AssemblyResolutionException available on core

sbomer:
We'd like to use this from the linker's assembly resolver.


jbevain:
Thanks!

PR #342, opened 2017-03-02
Fixed DocCommentId cast for `OptionalModifier`

joelmartinez:
An example ... for the following method in the `System.Printing.dll` assembly that ships with .net 4.6.2:

`System.ValueType modopt(System.DateTime) modopt(System.Runtime.CompilerServices.IsBoxed) System.Printing.IndexedProperties.PrintDateTimeProperty::op_Implicit(System.Printing.IndexedProperties.PrintDateTimeProperty)`

The return type's `MetadataType` is `OptionalModifier` ... and this code was original trying to cast it to `RequiredModifierType`.

### DIFF
--- a/rocks/Mono.Cecil.Rocks/DocCommentId.cs
+++ b/rocks/Mono.Cecil.Rocks/DocCommentId.cs
@@ -111,7 +111,7 @@ namespace Mono.Cecil.Rocks {
 					id.Append (((GenericParameter) type).Position);
 					break;
 				case MetadataType.OptionalModifier:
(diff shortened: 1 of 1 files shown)


joelmartinez:
I found this issue investigating a [runtime error in mdoc](https://github.com/mono/api-doc-tools/issues/33). 

@jbevain Please let me know if there are any other requirements for getting a PR merged (unit tests, etc.) and if so, how I can get this PR up to snuff. Thanks!

jbevain:
Thanks for the fix!

PR #343, opened 2017-03-02
Small method body caching.

### LINKED ISSUE #341: Cache method bodies for small methods
Roslyn has an optimization that reuses identical method bodies for small methods:
https://github.com/dotnet/roslyn/blob/91ee274439b6cbed8dbc42dccf3b737305652434/src/Compilers/Core/Portable/PEWriter/MetadataWriter.cs#L2875

This can be ported to Cecil.

Also I noticed that all method bodies are 4-byte aligned; that's only required for method bodies with fat headers, tiny method bodies don't have to be aligned.

erozenfeld:
This change implements small body caching: identical bodies for small methods are emitted only once.

The implementation is similar to Roslyn's implementation:
https://github.com/dotnet/roslyn/blob/91ee274439b6cbed8dbc42dccf3b737305652434/src/Compilers/Core/Portable/PEWriter/MetadataWriter.cs#L2875

Additionally, this change removes 4-byte alignment for tiny method body headers.
Only fat method body headers have to be 4-byte aligned.

Closes #341

### DIFF
--- a/Mono.Cecil.Cil/CodeWriter.cs
+++ b/Mono.Cecil.Cil/CodeWriter.cs
@@ -27,38 +27,35 @@ namespace Mono.Cecil.Cil {
 		readonly RVA code_base;
 		internal readonly MetadataBuilder metadata;
 		readonly Dictionary<uint, MetadataToken> standalone_signatures;
+		readonly Dictionary<byte [], RVA> small_method_bodies;
 
-		RVA current;
 		MethodBody body;
 
 		public CodeWriter (MetadataBuilder metadata)
 			: base (0)
 		{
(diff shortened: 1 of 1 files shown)


erozenfeld:
@jbevain PTAL

erozenfeld:
Looks like tests are failing because this code assumes 4-byte alignment for all body headers:
https://github.com/mono/mono/blob/master/mono/metadata/metadata-verify.c

erozenfeld:
From ECMA-335, II.25.4: Tiny method headers can start on any byte boundary.
Fat method headers shall start on a 4-byte boundary.

erozenfeld:
I looked at the test failures again and it turns out they were caused by lack of 4-byte alignment for unresolved method bodies. I fixed that and the tests are passing now.

PR #344, opened 2017-03-06
Fix SequencePoint.IsHidden comparing start_column, not end_line

0x0ade:
According to the [specification](https://github.com/dotnet/corefx/blob/master/src/System.Reflection.Metadata/specs/PortablePdb-Metadata.md) for the Portable PDB format,

> Hidden sequence point is a sequence point whose Start Line = End Line = 0xfeefee and Start Column = End Column = 0.

That also seems to apply to standard PDBs, at least those I've seen.

This commit only replaces the second half of the existing check, which checked the start column against `0xfeefee`.

Edit: I've admittedly forgotten the tests, yet am wondering why exactly it's failing.

### DIFF
--- a/Mono.Cecil.Cil/SequencePoint.cs
+++ b/Mono.Cecil.Cil/SequencePoint.cs
@@ -47,7 +47,7 @@ namespace Mono.Cecil.Cil {
 		}
 
 		public bool IsHidden {
-			get { return start_line == 0xfeefee && start_line == start_column; }
+			get { return start_line == 0xfeefee && start_line == end_line; }
 		}
 
 		public Document Document {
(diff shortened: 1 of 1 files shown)


jbevain:
That looks good but we have one test failing. Could you get this green before we merge? Thanks!

0x0ade:
I've made `IsHidden` take the previous behavior into account, meaning it now returns true if `Start Line = End Line = 0xfeefee` or `Start Line = Start Column = 0xfeefee`.

I still don't know why exactly the test case seems to differ from what I've seen so far or even what the spec seems to describe. I also fear that there may be more such PDBs in the wild.

The test now passes, although this honestly feels more like a cheap fix to make the test pass than a correct fix to take the root cause into account...

jbevain:
@0x0ade you're right, this is hiding the proper fix. I've implemented this in #344. Thank you for the initial work and raising this to my attention.

PR #345, opened 2017-03-08
Found a little dead code

zebmason:
Still working on my static analysis tool and looks like it found some dead code that you may wish to remove from the master branch


jbevain:
Thanks for this contribution! Just like last time, I won't merge the PR as is but I'll look into what it makes sense to pick. Please also do not submit a PR again for code that I voluntarily not touched :)

zebmason:
JB, your welcome. I've successfully removed all the false positives that my tool was reporting on Cecil leaving just Mono.Cecil.Tests.TypeDefinitionRocksTests.Foo.Baz() and 454 members in Microsoft.Cci.Pdb. Obviously didn't want to trouble you with those. Did consider trying to do some class splitting to improve cohesion but wasn't quite brave enough to do so without a better understanding of your code.

PR #346, opened 2017-03-09
Bring the ReflectionImporter to NET_CORE

jbevain:
The only remaining issue is getting a field definition/method definition from an instantiated type, see:

https://github.com/jbevain/cecil/blob/f4ee759024783e5f860b84d82499d70e03d48154/Mono.Cecil/Import.cs#L359-L375

### DIFF
--- a/Mono.Cecil.Cil/Symbols.cs
+++ b/Mono.Cecil.Cil/Symbols.cs
@@ -689,7 +689,7 @@ namespace Mono.Cecil.Cil {
 
 			var suffix = GetSymbolNamespace (kind);
 
-			var cecil_name = typeof (SymbolProvider).GetAssembly ().GetName ();
+			var cecil_name = typeof (SymbolProvider).Assembly ().GetName ();
 
 			var name = new SR.AssemblyName {
 				Name = cecil_name.Name + "." + suffix,
@@ -740,7 +740,7 @@ namespace Mono.Cecil.Cil {
 		static string GetSymbolTypeName (SymbolKind kind, string name)
 		{
 			var ns = GetSymbolNamespace (kind);
-			return typeof (SymbolProvider).GetAssembly ().GetName ().Name + "." + ns + "." + kind + name;
+			return typeof (SymbolProvider).Assembly ().GetName ().Name + "." + ns + "." + kind + name;
 		}
 
 		static string GetSymbolNamespace (SymbolKind kind)
--- a/Mono.Cecil/Import.cs
+++ b/Mono.Cecil/Import.cs
@@ -29,7 +29,7 @@ namespace Mono.Cecil {
 		MethodReference ImportReference (MethodReference method, IGenericParameterProvider context);
 	}
 
-#if !PCL && !NET_CORE
+#if !PCL
 
 	public interface IReflectionImporterProvider {
 		IReflectionImporter GetReflectionImporter (ModuleDefinition module);
@@ -126,7 +126,7 @@ namespace Mono.Cecil {
 		}
 	}
 
-#if !PCL && !NET_CORE
+#if !PCL
 
 	public class ReflectionImporter : IReflectionImporter {
 
@@ -159,7 +159,9 @@ namespace Mono.Cecil {
 			{ typeof (float), ElementType.R4 },
 			{ typeof (double), ElementType.R8 },
 			{ typeof (string), ElementType.String },
+#if !NET_CORE
 			{ typeof (TypedReference), ElementType.TypedByRef },
+#endif
 			{ typeof (IntPtr), ElementType.I },
 			{ typeof (UIntPtr), ElementType.U },
 			{ typeof (object), ElementType.Object },
@@ -179,8 +181,8 @@ namespace Mono.Cecil {
 				string.Empty,
 				type.Name,
 				module,
-				ImportScope (type.Assembly),
-				type.IsValueType);
+				ImportScope (type.Assembly ()),
+				type.IsValueType ());
 
 			reference.etype = ImportElementType (type);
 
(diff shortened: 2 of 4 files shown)


praeclarum:
Looks like just the CI broke? Is there anything I can do to help this along to release?

jbevain:
Yeah Travis went bonkers.
Have you tried it? Does it work for you?






On Fri, Mar 10, 2017 at 12:00 PM -0800, "Frank A. Krueger" <notifications@github.com> wrote:












Looks like just the CI broke? Is there anything I can do to help this along to release?

praeclarum:
No, I haven't will try right now.

praeclarum:
I'm having trouble building it.

When I select netstandard_Release config, VS complains that there aren't Build tasks for it. xbuild doesn't compile anything. Xamarin Studio grays out most of the projects and doesn't do anything when I say Build.

Which made me notice: you don't include Cecil.Rocks in the .NET Standard version. Is it possible to get at least the convenience methods in there?

When I select PCL, the methods I need aren't included.

jbevain:
The netstandard build requires the VS 2017 tooling and can be built using:

```
dotnet restore Mono.Cecil.csproj /p:Configuration=netstandard_Release
dotnet build Mono.Cecil.csproj /p:Configuration=netstandard_Release
```

As for Cecil.Rocks, I haven't had a look at building it on netstandard. Maybe just take the sources you need?

praeclarum:
Great. Should that work without VS2017 if I just have the latest .NET Core? Giving it a shot...

jbevain:
I think it should :)

praeclarum:
Compiled, now checking if this puppy runs on iOS.

praeclarum:
For the most part everything is working!

But I have some notes: ;-)

1. Would you consider including `DefaultAssemblyResolver`? It went in without changes but is still guarded by a define.

2. Most of Rocks went in fine too. I hope you'll consider adding these to Standard.

3. There is a change that is causing me problems. Assembly method bodies are not being loaded when I read the assembly. I load with:

```fsharp
let rparams = new Mono.Cecil.ReaderParameters (Mono.Cecil.ReadingMode.Immediate)
rparams.AssemblyResolver <- assemblyResolver resolve
let casm = Mono.Cecil.AssemblyDefinition.ReadAssembly (peStream, rparams)
```

But I get a runtime exception complaining: 

```
System.ObjectDisposedException: Cannot access a closed Stream.
  at System.IO.MemoryStream.get_Position () [0x0000b] in /Library/Frameworks/Xamarin.iOS.framework/Versions/10.4.0.123/src/mono/mcs/class/referencesource/mscorlib/system/io/memorystream.cs:322 
  at Mono.Cecil.PE.BinaryStreamReader.get_Position () [0x00006] in <bcd1ed9a158d4e2cb3deddc377322002>:0 
  at Mono.Cecil.Cil.CodeReader.MoveTo (Mono.Cecil.MethodDefinition method) [0x00013] in <bcd1ed9a158d4e2cb3deddc377322002>:0 
  at Mono.Cecil.Cil.CodeReader.ReadMethodBody (Mono.Cecil.MethodDefinition method) [0x00000] in <bcd1ed9a158d4e2cb3deddc377322002>:0 
  at Mono.Cecil.MetadataReader.ReadMethodBody (Mono.Cecil.MethodDefinition method) [0x00000] in <bcd1ed9a158d4e2cb3deddc377322002>:0 
  at Mono.Cecil.MethodDefinition+<>c.<get_Body>b__41_0 (Mono.Cecil.MethodDefinition method, Mono.Cecil.MetadataReader reader) [0x00000] in <bcd1ed9a158d4e2cb3deddc377322002>:0 
  at Mono.Cecil.ModuleDefinition.Read[TItem,TRet] (TRet& variable, TItem item, System.Func`3[T1,T2,TResult] read) [0x00040] in <bcd1ed9a158d4e2cb3deddc377322002>:0 
  at Mono.Cecil.MethodDefinition.get_Body () [0x0002c] in <bcd1ed9a158d4e2cb3deddc377322002>:0 
```

I thought that `Mono.Cecil.ReadingMode.Immediate` would let me close the stream and front load all the IO. I would prefer not to do any IO while interrogating method bodies. This code worked with the 0.9 series so I wonder if you added a new flag somewhere?

jbevain:
I think Immediate never read the method bodies but with 0.9 the assembly was stored in memory anyway.

I agree the current state feels weird. I'll think about it. The workaround is to trigger the read of all method bodies post read.

The DefaultAssemblyResolver only really makes sense if you want to resolve full framework assemblies. Which might make sense in netstandard. Are you resolving full framework assemblies? I was just considering adding a simple DefaultAssemblyResolver for netstandard that has the ability to lookup assemblies in directories.

Cecil.Rocks on netstandard: yeah definitely, no reason why not.

jbevain:
I assume Xamarin Forms needs the full DefaultAssemblyResolver when running on a full .NET profile.

@praeclarum: for your use case, is it safe to assume that all assemblies that Xamarin Forms would need to resolve are in the app directory?

praeclarum:
I have yet to try executing the Forms code on device yet, so I may end up just having to do very hacky things to the copy of DefaultAssemblyResolver I pulled in.

> is it safe to assume that all assemblies that Xamarin Forms would need to resolve are in the app directory?

On iOS it's weird now. With the simulator, all the assemblies turn into dylibs. On device, they all get linked into the main executable. So I end up having to using reflection to get at them.

praeclarum:
🎉 Happy to have this!

> I have yet to try executing the Forms code on device yet, so I may end up just having to do very hacky things to the copy of DefaultAssemblyResolver I pulled in.

Turns out that the one I pulled in does fine, no hacks needed.

My only gripe is that code using it doesn't resolve PCL types to their forwarded type (keeps System.Runtime types). But that most likely has nothing to do with Cecil.

jbevain:
> My only gripe is that code using it doesn't resolve PCL types to their forwarded type (keeps System.Runtime types). But that most likely has nothing to do with Cecil.

That's interesting, I would have thought that at runtime a System.Type would always point to its forwarded type.

praeclarum:
Yeah I'm confused by it all too. The assembly references X.F which contains a method returning a Func. When the code resolves Func, it gets the System.Runtime one (which X.F does reference).

I don't man, these type forwarders are such a hack...

PR #350, opened 2017-03-13
[Performance] Improve AddMapping performance by using a List instead …

Therzok:
…of an Array.

Mono.Cecil is used by mono-addins to scan all addin assemblies. In this
case, we request the types from a given assembly, which goes through the
nested types cache mapping.

A hot path in this case was AssemblyReader.AddMapping, which was doing a
lot of array resizing in the end.

Using mono-addins inside MonoDevelop would allocate 31MB of arrays.

MetadataReader.InitializeTypeDefinitions would allocate 57MB before, and
now it's down to 27MB.

The CPU time is also down now, from 117ms to 71ms.

### DIFF
--- a/Mono.Cecil/AssemblyReader.cs
+++ b/Mono.Cecil/AssemblyReader.cs
@@ -872,25 +872,25 @@ namespace Mono.Cecil {
 
 		public bool HasNestedTypes (TypeDefinition type)
 		{
-			uint [] mapping;
+			Collection<uint> mapping;
 			InitializeNestedTypes ();
 
 			if (!metadata.TryGetNestedTypeMapping (type, out mapping))
 				return false;
 
-			return mapping.Length > 0;
+			return mapping.Count > 0;
 		}
 
 		public Collection<TypeDefinition> ReadNestedTypes (TypeDefinition type)
 		{
 			InitializeNestedTypes ();
-			uint [] mapping;
+			Collection<uint> mapping;
 			if (!metadata.TryGetNestedTypeMapping (type, out mapping))
 				return new MemberDefinitionCollection<TypeDefinition> (type);
 
-			var nested_types = new MemberDefinitionCollection<TypeDefinition> (type, mapping.Length);
+			var nested_types = new MemberDefinitionCollection<TypeDefinition> (type, mapping.Count);
 
-			for (int i = 0; i < mapping.Length; i++) {
+			for (int i = 0; i < mapping.Count; i++) {
 				var nested_type = GetTypeDefinition (mapping [i]);
 
 				if (nested_type != null)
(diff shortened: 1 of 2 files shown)


Mono.Cecil/AssemblyReader.cs
```diff
@@ -929,18 +929,14 @@ void AddNestedMapping (uint declaring, uint nested)
 			metadata.SetReverseNestedTypeMapping (nested, declaring);
 		}
 
-		static TValue [] AddMapping<TKey, TValue> (Dictionary<TKey, TValue []> cache, TKey key, TValue value)
+		static List<TValue> AddMapping<TKey, TValue> (Dictionary<TKey, List<TValue>> cache, TKey key, TValue value)
 		{
```
Therzok: This is the only change that isn't a simple `T[]` -> `List<T>` and `Length` -> `Count`.

jbevain:
Thanks for having a look at this, the change looks good, thanks!
Could you just updated your change to Mono.Collections.Generic.Collection instead of List as it's used through Cecil.

For a future improvement, the point of having arrays here was to pay for the price of the allocations early to reduce the working set, but as you demonstrated it's slower and will make more short lived allocations. I'm wondering if we could not pre-compute the size of the mappings.

Jb

Therzok:
> I'm wondering if we could not pre-compute the size of the mappings.

That would mean a two-pass iteration, one to count, one to get the size itself. I could add timings for that and see how it goes.

> Could you just updated your change to Mono.Collections.Generic.Collection instead of List as it's used through Cecil.

I'll do that change.

Therzok:
I've updated it to use Collection instead of List. Pre-computing the size is not faster. It makes it have an overhead of a `Dictionary<uint, int>` allocation of size `length` and a double pass through the metadata. Slightly more allocations in this case (2MB) and some CPU increase to 86ms. I guess just using Collection is better in this case.

Therzok:
This was the tentative code for pre-computing:

```
void InitializeNestedTypes ()
		{
			if (metadata.NestedTypes != null)
				return;

			var length = MoveTo (Table.NestedClass);

			metadata.NestedTypes = new Dictionary<uint, Collection<uint>> (length);
			metadata.ReverseNestedTypes = new Dictionary<uint, uint> (length);

			if (length == 0)
				return;

			var countDict = new Dictionary<uint, int> (length);
			for (int i = 1; i <= length; i++) {
				var nested = ReadTableIndex (Table.TypeDef);
				var declaring = ReadTableIndex (Table.TypeDef);

				int old;
				countDict.TryGetValue (declaring, out old);
				countDict [declaring] = old + 1;
			}

			length = MoveTo (Table.NestedClass);

			for (int i = 1; i <= length; i++) {
				var nested = ReadTableIndex (Table.TypeDef);
				var declaring = ReadTableIndex (Table.TypeDef);

				AddNestedMapping (declaring, nested);
			}
		}
```

Didn't see a better way.

jbevain:
Let's go with what we have. Thanks for giving that a try. Could you fix up your second commit into the first one with rebase and force push to this branch before I merged? Thanks!

Therzok:
Done.

jbevain:
Another job well done! Thanks for your contribution!

PR #351, opened 2017-03-14
Native pdb roundtrip for scopes, async and using namespace

xen2:
PR for #90 

Readapted my old code to fit the new system, with some improvements.
Added native PDB roundtrip support for:
* Scopes
* Async debug info
* Using namespace/alias
* Fixed constant types/value (read only, write is not implemented yet)

Added some unit tests to check all of that is working well
BTW, I would be interested to have the source of `PdbTarget.cs` (right now I created my own `ComplexPdb.dll` with source alongside it in case somebody want to recompile it)


xen2:
Quick note, the other patch is to make `fixed` array work with an assembly that roundtrips through Cecil.

davidroth:
@SimonCropp Can you run the fody test suite on it?

jbevain:
This is awesome.

Thanks for working on this, I'll start a review.

jbevain:
@xen2 do you plan to push more commits to this PR short term?

I'm thinking that this PR is important and I don't want to spend too much time iterating back and forth on fixing up details. If you're done for now, I'll merge this into an integration branch where I'll do the cleanup and refactorings I'm thinking of.

xen2:
@jbevain Yes, pretty much done, nothing more short term.
I will try to fix the constant roundtrip at some point later but this is quite minor.

Testing it on real apps to make sure it works well enough (async stepping, namespace, etc...), but so far so good.

Happy this can get merged quickly!

xen2:
PS: Just forced pushed again (fixed some spaces to tabs)

jbevain:
Thank you for for your contribution, I'll do the integration work in #353!

PR #353, opened 2017-03-14
Native pdb integration

jbevain:
Continuation of #351 and preparation for master insertion.

- [X] Code style fixes
- [X] Refactoring redundant code
- [x] Fix native pdb constant roundtrip
- [x] Code reader changes


Mono.Cecil/AssemblyWriter.cs
```diff
@@ -3090,7 +3090,7 @@ static string GetXmlSecurityDeclaration (SecurityDeclaration declaration)
 
 		void WriteTypeReference (TypeReference type)
 		{
-			WriteUTF8String (TypeParser.ToParseable (type));
+			WriteUTF8String (TypeParser.ToParseable (type, false));
```
jbevain: @xen2 what's the reason for this?

xen2: Mentionned in the PR (sorry the commit message wasn't very explicit as to why):

> Quick note, the other patch is to make fixed array work with an assembly that roundtrips through Cecil.

Basically if mscorlib full name is not stored in the `FixedBufferAttribute` used for stuff such as `public fixed int[4]`, the library exposing this can't be consumed (both csc compiler and intellisense complains).

xen2: Note: also had to use the same trick in `NativePdbWriter` when importing `using static` which refers to a type, PDB encode it with full qualification even for mscorlib.

xen2:
Just so you know, the reason I made `AsyncMethodBodyDebugInformation(int)` ctor public is because in ILRepacker I sometime need to create one with offset -1 (which means no catch handler).

jbevain:
@xen2 good to know, we'll just need to expose this differently. The offsets constructors are meant to be internal only.

jbevain:
The WriteType / TypeParser changes are worrying me, they're basically bypassing the existing tests we have for TypeParser. I'll take them out of this PR to land this in master and work on them in a separate PR.

jbevain:
@xen2 thanks for your contribution!

xen2:
Agree that the TypeParser stuff is better separate and probably deserves its own test.
Thanks for sorting it out (sorry that my original branch included a few other somewhat unrelated changes) and merging it!

PR #354, opened 2017-03-14
Recognize System.Private.CoreLib as a core library.

erozenfeld:
@jbevain Let me know if you would prefer to have these lines under #if NET_CORE.


jbevain:
That looks about right without ifdefing, you might want to analyze such an assembly outside of .NET Core, and expect it to work with a proper resolver.

jbevain:
Thank you!

PR #355, opened 2017-03-15
Symbol writer provider external

xen2:
Objective of this commit is to avoid having to write this kind of code when I want to create a new assembly that keeps same symbol writer as another one, which will break as soon as there is a new symbol provider:
```
            ISymbolWriterProvider symbolWriterProvider = null;
            if (PrimaryAssemblyMainModule.SymbolReader is MdbReader)
                symbolWriterProvider = new MdbWriterProvider();
            else if (PrimaryAssemblyMainModule.SymbolReader is PortablePdbReader)
                symbolWriterProvider = new PortablePdbWriterProvider();
            else if (PrimaryAssemblyMainModule.SymbolReader is NativePdbReader)
                symbolWriterProvider = new NativePdbWriterProvider();
```

### DIFF
--- a/Mono.Cecil.Cil/Symbols.cs
+++ b/Mono.Cecil.Cil/Symbols.cs
@@ -873,17 +873,35 @@ namespace Mono.Cecil.Cil {
 
 #if !PCL
 	public class DefaultSymbolWriterProvider : ISymbolWriterProvider {
+		SymbolKind? default_reader_kind;
+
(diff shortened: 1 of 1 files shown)


jbevain:
Thanks for the PR. I'm closing this in favor of #358 which fixes the ugly string based detection I've added previously in SymbolProvider.

xen2:
Even better! 👍

PR #356, opened 2017-03-15
Open file with ReadWrite share mode if ReaderParameters.ReadWrite is true

xen2:
Had cases where I wanted to open an assembly in ReadWrite and also open it as a separate Stream in another thread (and ideally want to avoid passing stream around) or process.

### DIFF
--- a/Mono.Cecil/ModuleDefinition.cs
+++ b/Mono.Cecil/ModuleDefinition.cs
@@ -1100,7 +1100,7 @@ namespace Mono.Cecil {
 
 		public static ModuleDefinition ReadModule (string fileName, ReaderParameters parameters)
 		{
-			var stream = GetFileStream (fileName, FileMode.Open, parameters.ReadWrite ? FileAccess.ReadWrite : FileAccess.Read, FileShare.Read);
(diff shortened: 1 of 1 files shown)


jbevain:
I thought more about this and this feels odd. Why would you open another stream that could write the the stream where Cecil is reading and writing the assembly? Reading, why not, but writing?

xen2:
Yes, as mentionned in the other PR, not totally sure anymore either.
I will give it another try and let you know soon.

xen2:
Ok, please revert the commit.
Tried it again:
```
// This doesn't work (hence why I thought I needed FileShare.ReadWrite on both (so that the writer allow other to open)
File.Open(inputFile, FileMode.Open, FileAccess.ReadWrite, FileShare.Read);
File.Open(inputFile, FileMode.Open, FileAccess.Read, FileShare.Read);

// Turns out, this is enough:
File.Open(inputFile, FileMode.Open, FileAccess.ReadWrite, FileShare.Read);
File.Open(inputFile, FileMode.Open, FileAccess.Read, FileShare.ReadWrite); // ReadWrite here only
```

Note: The other PR is still useful so that Cecil can write to a file even though it is still being read by another external process.

jbevain:
Will do, thanks. For the other PR I'm considering just passing FileShare.Read in all cases.

xen2:
Sounds good, would probably avoid lot of unexpected `IOException`.

PR #357, opened 2017-03-15
FileShare can be customized when writing back the assembly

xen2:
Notes:
-  could also become a parameter in WriterParameters, but that wouldn't make sense when saving back a Stream, only with the fileName override when Cecil creates the stream -- let me know what you prefer
- Ideally, I could give an already opened stream but most `ISymbolWriterProvider.GetSymbolWriter(Stream)` are unimplemented. Of course, it is not easy because we would either want to provide user facility to determine SymbolStream filename on his own (to avoid hardcoding symbol file probing), and/or (as a fallback?) detect if Stream is a FileStream and fallback to file version?

### DIFF
--- a/Mono.Cecil/AssemblyDefinition.cs
+++ b/Mono.Cecil/AssemblyDefinition.cs
@@ -163,9 +163,9 @@ namespace Mono.Cecil {
 			Write (fileName, new WriterParameters ());
 		}
 
-		public void Write (string fileName, WriterParameters parameters)
+		public void Write (string fileName, WriterParameters parameters, FileShare share = FileShare.None)
 		{
-			main_module.Write (fileName, parameters);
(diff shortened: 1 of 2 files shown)


xen2:
Turns out I needed FileShare.Read (not ReadWrite). If I didn't set it, I often had IOException.

Tried to figure out with Process Monitor, and so far my best bet is AntiVirus and Windows Search that open the .exe file with FileShare.Read mode.
I think it happens because the file has just been written to and some other system process like this open it right away and still have it open by the time I try to save it again the second time (I write to the file twice in a row).

If Cecil tries to open it with FileShare.None, it will just throw an exception in that case.

jbevain:
I don't mind making it FileShare.Read for all the use cases in that case.

PR #361, opened 2017-03-21
Allow creation of AsyncMethodBodyDebugInformation with a null catch handler (encoded as offset -1 internally)

xen2:
Would it be OK that way?


jbevain:
I think it would be better if we had a parameterless constructor for this case. Also could you please follow the Mono coding guidelines, this way I could merge this as is without fixing it up :) Thanks!

xen2:
OK, force pushed with your proposed change.
Sorry for the Mono coding guideline, VS is fighting against me :)

Maybe I could commit a Resharper setting file that have those rules out of the box?

jbevain:
Thanks! There's a DotSettings for R# in the root folder.

PR #362, opened 2017-03-21
Recognize netstandard as a core library.

erozenfeld:


erozenfeld:
@jbevain PTAL

ericstj:
/cc @terrajobst @weshaggard
LGTM

jbevain:
Thanks, I fixed it slightly differently in the commit mentioned above.

terrajobst:
How do you detect the core assembly? Lookup by names have a tendency to be fragile and not to work very well when you inspect mixed-cases that involves facades. Ideally, you'd detect it by assuming the core assembly is the one without dependencies.

jbevain:
The diff mentioned previously has all the details: Name + empty AssemblyRef table.
Jb, on a phone.
		

		
	_____________________________

PR #363, opened 2017-03-21
Make Mono.Cecil.Pdb compile on netstandard

jbevain:


erozenfeld:
LGTM. Thanks for fixing this!

PR #368, opened 2017-03-29
[ImageWriter] zero stream length before writting

radekdoulik:
- it fixes the case, where we have an assembly read with
   ReadWrite=true parameter, remove parts of it and write it
   back. before we ended with the new assembly file with the same size
   as the original, so it wasn't truncated


jbevain:
Thanks, I've fixed this slightly differently. This will go in beta 6.

PR #369, opened 2017-03-29
Enable restoring of Cecil as a P2P reference in .NET Core

sbomer:
A NuGet bug (https://github.com/NuGet/Home/issues/4873) causes the
configuration not to be set during restore when using Cecil as a
ProjectReference of another project using the .NET Core cli tooling.

Until that bug is fixed, this workaround will correctly import
NetStandard.props by checking for the existence of an msbuild property
that is set during restore using the new tooling.

### DIFF
--- a/Mono.Cecil.csproj
+++ b/Mono.Cecil.csproj
@@ -16,5 +16,5 @@
     <Compile Include="Mono.Security.Cryptography\*.cs" />
   </ItemGroup>
   <Import Project="Mono.Cecil.props" />
-  <Import Project="$(MSBuildCSharpTargets)" Condition=" ! $(Configuration.StartsWith('netstandard')) " />
+  <Import Project="$(MSBuildCSharpTargets)" Condition=" ! $(NetStandard) " />
 </Project>
--- a/Mono.Cecil.props
+++ b/Mono.Cecil.props
@@ -12,6 +12,8 @@
     <OutputType>Library</OutputType>
     <OutputPath>$(BuildDirectory)\bin\$(Configuration)\</OutputPath>
     <MSBuildCSharpTargets>$(MSBuildToolsPath)\Microsoft.CSharp.targets</MSBuildCSharpTargets>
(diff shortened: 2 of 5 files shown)


jbevain:
A few things:

* Instead of having complicated conditions, we should just set a property in Mono.Cecil.props for a NetStandard build. I suggest setting a property NetStandard to true depending on the configuration and this NuGetRestoreTargets property, and using that in conditions.
* Now that Mono.Cecil.Mdb, Mono.Cecil.Pdb and Mono.Cecil.Rocks are built on netstandard as well, their respective csproj need to be updated as well.

Thanks!

sbomer:
@jbevain I made the suggested changes.

jbevain:
@sbomer thanks. Sorry if I wasn't clear, I'd like to keep the Import under the NetStandard condition into the csproj. Please address this, then fixup your changes into 1 commit and this will be good to go!

Thanks!

sbomer:
Oops, just realized that I moved the imports around - I'll fix this in a bit.

sbomer:
@jbevain done. I think this is what you wanted.

jbevain:
Ahah, one more for the road: you should be able to say `Condition=" ! $(NetStandard) "` :)

sbomer:
If `$(NetStandard)` isn't set to `true` it will be empty, and this will cause errors like this during the build:
`Specified condition " ! $(NetStandard) " evaluates to "!" instead of a boolean.`

Using `== ''` and `!= ''` seems to be a common pattern in msbuild files. The alternative would be to insert something like
`<NetStandard Condition=" '$(NetStandard)' == '' ">false</NetStandard>`
If you prefer that, let me know and I'll make the change. :)

jbevain:
Yeah I'm thinking at this point it should be a boolean that we can query, so let's make the change and get this merged :)

sbomer:
@jbevain done

jbevain:
Thanks!

sbomer:
Thank you!

PR #373, opened 2017-04-06
Upstream dont call windowsruntime during read

joshpeterson:
We found a crash deep in the metadata reading code when Windows Runtime projections are calculated before the type is fully read. This showed up with the latest Xbox One SDK.

### DIFF
--- a/Mono.Cecil/AssemblyReader.cs
+++ b/Mono.Cecil/AssemblyReader.cs
@@ -1,4 +1,4 @@
-//
+﻿//
 // Author:
 //   Jb Evain (jbevain@gmail.com)
 //
@@ -837,6 +837,12 @@ namespace Mono.Cecil {
 
 				types [i] = ReadType (i + 1);
 			}
+
+			if (module.IsWindowsMetadata()) {
+				for (uint i = 0; i < length; i++) {
+					WindowsRuntimeProjections.Project (types [i]);
+				}
+			}
 		}
 
 		static bool IsNested (TypeAttributes attributes)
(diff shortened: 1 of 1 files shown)


Mono.Cecil/AssemblyReader.cs
```diff
@@ -1,4 +1,4 @@
-//
+﻿//
```
jbevain: Not necessary :)

Mono.Cecil/AssemblyReader.cs
```diff
@@ -837,6 +837,12 @@ void InitializeTypeDefinitions ()
 
 				types [i] = ReadType (i + 1);
 			}
+
+			if (module.IsWindowsMetadata()) {
```
jbevain: Missing space before parenthesis: `IsWindowsMetadata ()`

Mono.Cecil/AssemblyReader.cs
```diff
@@ -1045,7 +1048,13 @@ public TypeDefinition GetTypeDefinition (uint rid)
 			if (type != null)
 				return type;
 
-			return ReadTypeDefinition (rid);
+			type = ReadTypeDefinition (rid);
+
+			if (module.IsWindowsMetadata()) {
```
jbevain: Missing space before parenthesis. Also please remove braces here.

joshpeterson:
Where would you recommend tests for this code go? For example, I don't see any direct calls to `GetTypeDefinition` from the unit test code, but I do see it called indirectly from a number of tests.

jbevain:
One integration test that would previously trigger the crash you mention would be great. They're in:

https://github.com/jbevain/cecil/blob/master/Test/Mono.Cecil.Tests/WindowsRuntimeProjectionsTests.cs

joshpeterson:
Thanks. I'd like to add a .winmd file that reproduces this then, but honestly I'm unsure how to reproduce it. We noticed it in a file from the Xbox One SDK, but we cannot commit that as it is under NDA. :) Do you have any idea about what might cause the issue?

jbevain:
Oh well :)

joshpeterson:
Sorry, I completely forgot about this PR!

Issue #372, opened 2017-04-06
Support not putting the full path to the pdb in assemblies


rolfbjarne:
Cecil writes the full path to the pdb here: https://github.com/jbevain/cecil/blob/7297b51/Mono.Cecil.Cil/PortablePdb.cs#L304

Supporting a way to only write the filename would fix this: https://bugzilla.xamarin.com/show_bug.cgi?id=54578

Roslyn has a similar issue: https://github.com/dotnet/roslyn/issues/9813

jbevain:
Fixed by #380.

PR #380, opened 2017-04-18
Only write down the filename of the portable pdb in the debug header

jbevain:


jbevain:
Writing only the file name is a sane default and should fix #372.

jbevain:
cc @spouliot @rolfbjarne.

PR #381, opened 2017-04-20
Dispose stream in WriteModuleTo on .NET Core

sbomer:
This fixes a bug I was encountering on windows in which a subsequent attempt to open the file in the same process failed.


jbevain:
Oops. Thanks!

sbomer:
@marek-safar, could you get this into mono/cecil?

PR #384, opened 2017-04-25
Release lock on file when exception in ModuleWriter.WriteModuleTo

kswoll:
Currently, if an exception occurs in `ModuleWriter.WriteModuleTo` and the `Disposable<Stream>` is "owned", `Dispose()` will not be called, and the file will be locked.  

For example, in a `Fody` plugin, if this situation arises, then one must manually kill `MsBuild.exe` in order to release the lock.  I think probably `Dispose()` ought to be called in all circumstances.  (Currently, an exception can arise with Fody if you take a loaded `MethodDefinition` for a method that contains a local variable and assign it a new `MethodBody` the cached debug info becomes invalid and an exception will occur.   But I'll file a separate PR for that)

* Add a unit test verifying the problem/fix
* Surround method body with a try/finally

### DIFF
--- a/Test/Mono.Cecil.Tests/ModuleTests.cs
+++ b/Test/Mono.Cecil.Tests/ModuleTests.cs
@@ -294,5 +294,28 @@ namespace Mono.Cecil.Tests {
 			using (var module = ModuleDefinition.ReadModule (path))
 				Assert.AreEqual ("Foo.Foo", module.Types [1].FullName);
 		}
+
+		[Test]
+		public void ExceptionInWriteDoesNotKeepLockOnFile ()
+		{
+			var path = Path.GetTempFileName ();
+
+			var original = ModuleDefinition.CreateModule ("FooFoo", ModuleKind.Dll);
+			var type = new TypeDefinition ("Foo", "Foo", TypeAttributes.Abstract | TypeAttributes.Sealed);
+			original.Types.Add (type);
+			original.Attributes = 0;      // Force an exception
+			try
+			{
+				original.Write(path);
+				Assert.Fail("Exception should have been thrown");
+			}
+			catch
+			{
+				// Expected
+			}
+
+			// Ensure you can still delete the file
+			File.Delete (path);
+		}
 	}
 }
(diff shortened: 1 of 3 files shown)


jbevain:
Good catch, thanks!

Test/Mono.Cecil.Tests/ModuleTests.cs
```diff
@@ -294,5 +294,28 @@ public void ReadAndWriteFile ()
 			using (var module = ModuleDefinition.ReadModule (path))
 				Assert.AreEqual ("Foo.Foo", module.Types [1].FullName);
 		}
+
+		[Test]
+		public void ExceptionInWriteDoesNotKeepLockOnFile ()
+		{
+			var path = Path.GetTempFileName ();
+
+			var original = ModuleDefinition.CreateModule ("FooFoo", ModuleKind.Dll);
```
SimonCropp: the caller of ModuleDefinition.CreateModule is responsible for disposing of it? does the test still fail with that dispose?

jbevain: The module `original` here is freshly created and doesn't hold to a backing `Stream`.
Calling `Write(string)` means Cecil creates and owns the stream to write to. Cecil crashes, and leaves the stream open while it failed to write to it.

I think the issue is valid.

SimonCropp: @jbevain yep sorry missed that.

SimonCropp:
@jbevain if u ping when u merge this i will quickly update fody and do a smoke test for u

jbevain:
This PR is a great find.

We have the same issue with symbol providers where we need to ensure they get disposed.

kswoll:
@jbevain Awesome, happy to have helped in a small way! :)

jbevain:
Fixed it slightly differently in #385. Thanks again for reporting this!

kswoll:
Awesome, thanks for the quick turnaround!  cc @SimonCropp

kswoll:
@SimonCropp just curious if you've had a chance to come back to this?  Our team would prefer to wait until fody is updated with this before upgrading to 2.0.  Thanks!

SimonCropp:
@kswoll this should now be on nuget

kswoll:
Thanks @SimonCropp!

PR #387, opened 2017-05-01
Delay-sign cecil assemblies for .NET Core build

sbomer:
Delay-signing cecil is necessary so that we can authenticode-sign cecil for .NET core as part of the linker build. I'm not sure if this is an acceptable default for the cecil project, so please let me know if there's a better way to do this. Ideally I would like to isolate cecil's project files from the changes necessary in the linker build, but I don't see a great way to do this with submodules.

An alternative to this would be to enable authenticode signing in the cecil project, but that would probably involve even more intrusive changes to cecil.

Thoughts?

### DIFF
--- a/NetStandard.props
+++ b/NetStandard.props
@@ -5,5 +5,6 @@
     <EnableDefaultCompileItems>false</EnableDefaultCompileItems>
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
     <PublicSign Condition=" '$(OS)' != 'Windows_NT' ">true</PublicSign>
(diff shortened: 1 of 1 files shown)


sawilde:
I thought that strong names and authenticode were two different things - if you authenticode signed the delay signed builds and then completed the string name signing wouldn't this break the authenticode signing?

sbomer:
You're right - they're different, and our tool will do both. We have some signing infrastructure that we use to first strong-name sign and then authenticode sign assemblies, and they complain if the inputs are already strong-named. That's my understanding. @chcosta may know more.

sbomer:
Perfect, thanks for the tip!

PR #388, opened 2017-05-04
Get readonly building again

SimonCropp:


Mono.Cecil.props
```diff
@@ -31,6 +31,22 @@
     <WarningLevel>4</WarningLevel>
     <RunCodeAnalysis>false</RunCodeAnalysis>
   </PropertyGroup>
+  <PropertyGroup Condition=" $(Configuration.EndsWith('Debug_ReadOnly')) ">
```
jbevain: Maybe change the previous property group condition to Contains('Debug') and Contains('Release') and add a property group Contains('ReadOnly') to avoid duplication?

jbevain:
Well that makes sense.

Mono.Cecil.props
SimonCropp: done

SimonCropp:
@jbevain fixed the duplication

SimonCropp:
What r your thoughts on shipping a readonly nuget? Would we need to change the dll names and namespaces?

jbevain:
I don't think it's worth changing the namespaces, but yeah, we'd need new assembly names.

Issue #390, opened 2017-05-04
Support writing pure IL for ready-to-run mixed-mode assemblies.


sbomer:
Since the netcoreapp packages are starting to ship with crossgen'd ready-to-run images, we need the [linker](https://github.com/mono/linker) to be able to work with R2R images. We'd like to be able to write back pure IL assemblies for linked R2R images.

Currently that doesn't work because of the check in [ModuleWriter.Write](https://github.com/jbevain/cecil/blob/007b6dcaaf5c52bb5b3efa658ba63b2aa5de2d2d/Mono.Cecil/AssemblyWriter.cs#L83):
```
if ((module.Attributes & ModuleAttributes.ILOnly) == 0)
	throw new NotSupportedException ("Writing mixed-mode assemblies is not supported");
```

If I work around that by setting ModuleAttributes.ILOnly from the linker, I hit the following:

```
System.NotSupportedException: Specified method is not supported.
   at Mono.Cecil.PE.ImageWriter.GetMachine()
   at Mono.Cecil.PE.ImageWriter.WritePEFileHeader()
   at Mono.Cecil.PE.ImageWriter.WriteImage()
   at Mono.Cecil.ModuleWriter.WriteModuleTo(ModuleDefinition module, Disposable`1 stream, WriterParameters parameters)
   at Mono.Cecil.ModuleDefinition.Write(String fileName, WriterParameters parameters)
   at Mono.Cecil.AssemblyDefinition.Write(String fileName, WriterParameters parameters)
```

This is because the machine field in the COFF header is set to `0xfd1d`, which isn't recognized by cecil. [GetMachine()](https://github.com/jbevain/cecil/blob/bbe18d77b98746bbc0057d91693760f76b0735d3/Mono.Cecil.PE/ImageWriter.cs#L214) doesn't recognize it as a valid [TargetArchitecture](https://github.com/jbevain/cecil/blob/bbe18d77b98746bbc0057d91693760f76b0735d3/Mono.Cecil/ModuleKind.cs#L28).

[...]

PR #391, opened 2017-05-04
Target arch fix

sbomer:
Clean up some code to use the TargetArchitecture enum values introduced in https://github.com/jbevain/cecil/commit/bb40c2108ecf303691d0536c4f9d3b9035790c5c, as suggested by @jbevain in https://github.com/jbevain/cecil/issues/390.


jbevain:
Thanks!

Issue #389, opened 2017-05-04
Expose comparison operations on TypeReference, MethodReference etc.


atykhyy:
Code involving complex assembly modification or generation often needs to compare `TypeReference`, `MethodReference` and similar types for equality and/or use them as dictionary keys. Currently this involves duplication of code already present and used in Mono.Cecil\MetadataResolver.cs. Would you consider exposing these "official" comparison methods? As I see it, there are two options: (1) put the existing static `AreSame` methods into a separate public class and make the minimum number of them public, or (2) have `TypeReference` and similar types implement `IEquatable<>`  (overriding `object.Equals/GetHashCode` would be a potentially breaking change and is probably a bad idea anyway). I can do a PR if you OK an approach.

jbevain:
Hey Anton,

There's work for the winmd support that needs something similar so I'd say yes, that's planned.

Thanks.

jbevain:
This will likely come as part of #394.

atykhyy:
Thanks for the update. The type and method reference comparers in #394 call Resolve() and compare definitions as the final step, which is a big difference from the reference comparers in MetadataResolver.cs. References may not always be resolvable to definitions - method/type not yet emitted, assembly not available etc. If I am not mistaken, the module writer currently writes type and method references without resolving them, which is a big help (and even a lifesaver) in some situations. I don't know anything about Windows Runtime projections, though.

PR #392, opened 2017-05-05
Expose XxxReference comparisons

atykhyy:
Re issue #389.

### DIFF
--- a/Mono.Cecil/Import.cs
+++ b/Mono.Cecil/Import.cs
@@ -301,14 +301,6 @@ namespace Mono.Cecil {
 			if (TryGetAssemblyNameReference (name, out scope))
 				return scope;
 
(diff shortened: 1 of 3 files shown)


AnakinSklavenwalker:
what's holding this one back from merging @jbevain ?

jbevain:
@AnakinSklavenwalker comparing metadata elements is tricky, and not everybody will want to do exactly the same thing, I'm not very excited about providing a public API that would become the source of truth for is a `TypeReference` equal to another `TypeReference`.

AnakinSklavenwalker:
fair enough. Would you be OK having the comparer into the Rocks project. This way it can be documented that this might not be a 100% correct implementation, (according to the ECMA spec) but it works for most cases. 
Since I need such a comparer for my project anyway, i could create a new PR which re-implements the feature and puts it into Rocks.

PR #394, opened 2017-05-05
Unity's Windows Runtime changes

TautvydasZilys:


mrvoorhe:
@jbevain are there any changes you want to see before you would land this?

jbevain:
@mrvoorhe @TautvydasZilys can't we all collectively pretend that winmd files don't exist?

mrvoorhe:
@jbevain you can.  Sometimes I can :smile: and @TautvydasZilys never can.

My end goal is to reduce our custom changes to cecil as much as possible so that it is easier for us to track upstream master.

@jbevain Based on your own comment https://github.com/jbevain/cecil/issues/389#issuecomment-303771487 there is at least something in this PR that you are open to.  Do you want us to break some pieces out of this PR?

jbevain:
I think reducing the difference between your version and upstream is a worthy goal.

I also fully trust Tautvydas. It's me the problem. Taking this means that I'll probably end up maintaining it from my retirement home, and I just don't understand anything related to those projections because I've never had to look into them. That makes me skittish to take such a large change, because in the end I have no idea what I'm taking.

jbevain:
The counter argument to that is we already landed part the support for winmd files, and that PR fixes it. So I might as well just take it already.

mrvoorhe:
It sounds like you are making progress sorting out your dilemma with yourself.  Keep it up!

mrvoorhe:
‘Tis the season to click green merge buttons

jbevain:
Thanks for the quick fix folks, I'm glad we've been able to work diligently to get this merged.

PR #405, opened 2017-06-02
Add support for source link and embedded sources custom debug infos

jbevain:


jbevain:
That will go in beta7.

PR #416, opened 2017-06-08
Allow for processor architecture when resolving

lkinsella:
One of the problems I've run into with the existing `BaseAssemblyResolver` is that it will resolve on a first-seen wins basis so that you can have 64-bit GAC assemblies being returned for pure 32-bit applications, and even the wrong mscorlib. Although on the surface they're the same usually, technically they're not.

To resolve this I've reworked significant portions of the `BaseAssemblyResolver` to now accept a target architecture. 

I have renamed the existing `TargetArchitecture` to `ProcessorArchitecture` which I would argue is what it should really be. I then introduced a new `TargetArchitecture` enum with the following values:

- `Any` - Allow anything, this is the default (and existing) behaviour.
- `ILOnly` - Resolve only pure IL assemblies.
- `x86` - Resolve x86 assemblies and IL assemblies.
- `x86Only` - Resolve only x86 assemblies.
- `x64` - Resolve x64 assemblies and IL assemblies.
- `x64Only` - Resolve only x64 assemblies.
- `ARM*` - Not currently processed, not quite sure what to do with these.

When resolving GAC based assemblies the resolver will now first collect all matching assemblies rather than just return the first. It will then run through this list of matches and filter out any incompatible architectures before returning the top result of the filtered set.

For individual assemblies I've also modified the `GetAssembly` method so that it runs through a similar filtering process and will return `null` if it cannot find a match.

You can set `TargetArchitecture` as part of the `ReaderParameters` class. 

I also ran into an issue with the way `mscorlib` is resolved. At present the filename is got using something like this:

```
var fileName = typeof(object).Module.FullyQualifiedName;
```

The problem with this is that the `mscorlib` is based on the executing process architecture and _not_ the actual architecture that's relevant to the resolution.

So I now try and get it from the GAC first (using all of the above), failing that and if another new propery on `ReaderParameters` called `AllowAnyCorlib` is `false`, will return `null`. By default however it's `true` to keep existing behaviour.

A concrete example of use:

```
if (Executable.Is64Bit)
{
	targetArchitecture = TargetArchitecture.x64;
}
else if (Executable.Is32Bit)
{
	targetArchitecture = TargetArchitecture.x86;
	
	if (Environment.Is64BitOperatingSystem && !Executable.IsCLR32BitPreferred && !Executable.IsCLR32BitRequired)
		targetArchitecture = TargetArchitecture.x64;
}

var readerParams = new ReaderParameters() {
	AssemblyResolver = new AssemblyResolver(this),
	AllowAnyCorlib = false,
	TargetArchitecture = targetArchitecture
};
var moduleDef = ModuleDefinition.ReadModule(Executable.Filename, readerParams);
```

### DIFF
--- a/Mono.Cecil.PE/Image.cs
+++ b/Mono.Cecil.PE/Image.cs
@@ -26,7 +26,7 @@ namespace Mono.Cecil.PE {
 
 		public ModuleKind Kind;
 		public string RuntimeVersion;
-		public TargetArchitecture Architecture;
+		public ProcessorArchitecture Architecture;
 		public ModuleCharacteristics Characteristics;
 
 		public ImageDebugHeader DebugHeader;
--- a/Mono.Cecil.PE/ImageReader.cs
+++ b/Mono.Cecil.PE/ImageReader.cs
@@ -92,9 +92,9 @@ namespace Mono.Cecil.PE {
 			image.Characteristics = (ModuleCharacteristics) dll_characteristics;
 		}
 
-		TargetArchitecture ReadArchitecture ()
+		ProcessorArchitecture ReadArchitecture ()
 		{
-			return (TargetArchitecture) ReadUInt16 ();
+			return (ProcessorArchitecture) ReadUInt16 ();
 		}
 
 		static ModuleKind GetModuleKind (ushort characteristics, ushort subsystem)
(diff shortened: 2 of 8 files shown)


jbevain:
Hey,

Thank you for contributing this and sorry for the time it took to review this.

I understand the reasons why you'd want this change, but unfortunately:

* the PR changes the formatting of the code a lot,
* this introduces 1 unnecessary API break,
* and a large change to the BaseAssemblyResolver which I'm wary about.

And I'm afraid I'm not going to take this in for now.

What I suggest is that you write your own IAssemblyResolver implementation, composing over or reusing the BaseAssemblyResolver and pass that to your modules as the AssemblyResolver to use.

Thanks!

PR #419, opened 2017-06-14
Don't stop looking for symbols if we can't load a native pdb symbol provider. Fixes #56808.

rolfbjarne:
We might run into native pdbs, but we might not include support for native
pdbs, in which case we shouldn't stop looking for other debug symbols we might
be able to load (mdbs).

https://bugzilla.xamarin.com/show_bug.cgi?id=56808


rolfbjarne:
I've opened an identical PR for mono/cecil#15.

jbevain:
@rolfbjarne makes sense. Could you wrap the call CreateInstall for the mdb provider as well?

rolfbjarne:
@jbevain done

jbevain:
Thanks Rolf!

PR #420, opened 2017-06-17
add readonly nuspec

SimonCropp:


SimonCropp:
@jbevain anything else u need on this one?

SimonCropp:
Gentle reminder about my PR ;)

jbevain:
I'm conflicted about this. I don't mind maintaining the readonly scenario from source, but I'm not sure we want to go as far as publishing a nuget. That always seemed more of a scenario where people either build Cecil from source or override the build with their custom assembly name anyway.

SimonCropp:
these days if it not on nuget it may as well not exist.

jbevain:
NuGet users can use the normal Cecil and won't miss anything. I think there's only a very small fringe of the population that's interested in shaving off a few bytes (which is only what ReadOnly does).

SimonCropp:
How about if i ask really nicely. Tbh i would like to use this in enough projects that i would manage my own readonly nuget

PR #424, opened 2017-06-22
fix reading scope start

radekdoulik:
- in case when the mdb file is out of sync, we might endup with
   start_instruction being null

 - it is regression introduced by
   7c8e0f767d7b2f652a430e4e26f1d98a20f9e125

### DIFF
--- a/Mono.Cecil.Cil/CodeReader.cs
+++ b/Mono.Cecil.Cil/CodeReader.cs
@@ -160,7 +160,8 @@ namespace Mono.Cecil.Cil {
 		void ReadScope (ScopeDebugInformation scope)
 		{
 			var start_instruction = GetInstruction (scope.Start.Offset);
-			scope.Start = new InstructionOffset (start_instruction);
+			if (start_instruction != null)
+				    scope.Start = new InstructionOffset (start_instruction);
 
 			var end_instruction = GetInstruction (scope.End.Offset);
 			scope.End = end_instruction != null


jbevain:
I'm not opposed to merging this as this is pretty benign, but by loading and out of sync mdb and letting Cecil create an object model you're opening the door to having all sorts random of error.

My advice would be to focus on making sure you can not open an out of sync mdb.

radekdoulik:
I agree it is better to not open such assemblies. Sometime it is hard to know in before, the assembly and mdb can be for example from nuget package.

So it will help to have this merged or change the behaviour so that it throws something else than System.ArgumentNullException. If that is the case then I think it should do the same for start_instruction and end_instruction.

radekdoulik:
I have updated the patch, because in previous version it was causing issues when writting back the mdb. The updated patch get us to the state of setting scope start we had before and it now works OK even when writing mdb.

I am now trying to find out whether that can be avoided earlier. I suspect that either mvid wasn't changed when the assembly with broken mdb was updated or mdb reader might not be checking the mvid.

jbevain:
The `MdbReader` does check the GUID:

https://github.com/jbevain/cecil/blob/24dec0becc6026f4c01377f20da5f0fe9950b892/symbols/mdb/Mono.Cecil.Mdb/MdbReader.cs#L70

As a rule of thumb, a tool processing assemblies with mdb symbols should make sure to renew the mvid at every rewrite.

radekdoulik:
I traced it down to Forms task XamlC, which alters the assembly without changing the MVID. These are the relevant bugs https://bugzilla.xamarin.com/show_bug.cgi?id=57931, https://bugzilla.xamarin.com/show_bug.cgi?id=57889.

radekdoulik:
It would be nice to have this PR merged so that we don't get just System.ArgumentNullException when that happens.

radekdoulik:
@jbevain what is your opinion on whether cecil should handle broken mdb's? should I look into another solution with throwing an exception indicating the symbols are out of sync? Like introducing new SymbolFileException?

jbevain:
@radekdoulik I'm conflicted on this. If we hide the exception, we're allowing Cecil to create a representation for what is essentially an invalid state, that will probably crash later on when writing. On the other hand, it's also very easy with Cecil to generate something invalid, so should we bother?

In the ideal solution, we could detect a mismatch from MdbReader.ProcessDebugHeader and return false.

radekdoulik:
@jbevain fortunately it doesn't crash on write, because the scope start offset still has its original int value read from the mdb file -
 https://github.com/jbevain/cecil/blob/1530a21d652992e9adb749d207a4f4fd44ad8024/Mono.Cecil.Cil/Symbols.cs#L183. That is how it worked for us before. We wrote the new mdb with the same broken values.

I see that it hides the other issues, on how we get the broken symbol file in the first place. Thus I think getting more informative exception might be an alternative and would help to handle the problem in cecil's client code like linker. We would at least know the problem is related to the symbol file and might try to re-run linker without reading the symbol file and print warning about it.

I am also not really sure what would be better. This PR solution would help us quicker and will return to the behaviour we depended on, the exception handling case might be cleaner.

jbevain:
Let's get this merged until we figure out the long run.

PR #429, opened 2017-07-06
Adding IMethodBody interface to allow different kinds of body implementations

edgardozoppi:
Hi,

This PR is just a proof-of-concept to see how difficult could be to allow different kinds of method body implementations besides the currently provided MethodBody class for IL in Cecil API and also to start a discussion about this.

The idea is to provide the user the possibility to reuse the Cecil metadata model while extending it by implementing his own representation of method bodies. This way other code representations can be used instead of CIL, allowing higher levels of abstraction. For example, Abstract Syntax Tree (AST), Three Address Code (TAC), Static Single Assignment form (SSA), etc. Is important to note that Cecil doesn't have to provide these implementations nor support them, just need to be flexible enough to allow users to do it themselves.

I just included the Method property in the IMethodBody interface to keep it simple for the moment and to avoid imposing too many members that all other implementations must provide. But I'm not entirely sure about this decision, maybe it makes sense to also include the following properties:

```csharp
bool HasVariables { get; }
Collection<VariableDefinition> Variables { get; }
ParameterDefinition ThisParameter { get; }
```

I have also extended the Mixin class to include a handy helper method AsILMethodBody to avoid having to cast manually in every location where the original (IL) MethodBody class implementation is expected.

Please let me know what do you think about this and I will be happy to update the PR with the requested changes.

Thanks,
Edgardo.

### DIFF
--- a/Mono.Cecil.Cil/CodeWriter.cs
+++ b/Mono.Cecil.Cil/CodeWriter.cs
@@ -50,7 +50,7 @@ namespace Mono.Cecil.Cil {
 
 				rva = WriteUnresolvedMethodBody (method);
 			} else {
-				if (IsEmptyMethodBody (method.Body))
+				if (IsEmptyMethodBody (method.Body.AsILMethodBody()))
 					return 0;
 
 				rva = WriteResolvedMethodBody (method);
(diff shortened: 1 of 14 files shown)


jbevain:
Hello!

Thanks for contributing this but unfortunately that's not something we're going to take right now.

The change is severely breaking existing users, for a benefit that is hard to measure. If you have an existing implementation for different method bodies that need this, I'll be happy to have a look.

PR #435, opened 2017-08-09
Include formatting style in .editorconfig

tmat:


tmat:
@jbevain Please review.

jbevain:
Thanks! Is there a list of settings supported by this format? It would be great to override most of the settings impacted by the Mono coding style.

PR #442, opened 2017-09-01
Update solution to the new project system

### LINKED ISSUE #407: new project system
would u accept a PR that converts the solution to the new 2017 csproj system?

Or perhaps u can just use the PR to inform how you would like to move over?

sharwell:
Closes #407


jbevain:
Thanks for contributing this.

I'm not opposed to taking this change, but we need to make sure to retain the override feature.

I also need to check with a few downstream users if it impacts them.

jbevain:
Paging @marek-safar @spouliot @rolfbjarne @TautvydasUnity @jonpryor @kzu @SimonCropp 

Any comment? If you are building from source, is this a problem for you?

Mono.Cecil.props
```diff
@@ -1,101 +0,0 @@
...
-using System.Reflection;
-using System.Runtime.CompilerServices;
-using System.Runtime.InteropServices;
-
-[assembly: InternalsVisibleTo ("MyCompany.Cecil.Pdb, PublicKey=....")]
-[assembly: InternalsVisibleTo ("MyCompany.Cecil.Mdb, PublicKey=...")]
-[assembly: InternalsVisibleTo ("MyCompany.Cecil.Rocks, PublicKey=...")]
-[assembly: InternalsVisibleTo ("MyCompany.Cecil.Tests, PublicKey=...")]	
-  -->
-</Project>
```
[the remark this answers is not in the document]
sharwell: ❓ How are we expecting to handle these users given the migration to the new project system?

NetStandard.props
```diff
@@ -1,9 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003" Sdk="Microsoft.NET.Sdk">
-  <PropertyGroup>
-    <DefineConstants>$(DefineConstants);NET_4_0;NET_CORE;</DefineConstants>
-    <EnableDefaultCompileItems>false</EnableDefaultCompileItems>
-    <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
-    <PublicSign Condition=" '$(OS)' != 'Windows_NT' ">true</PublicSign>
```
sharwell: ❓ This condition was dropped from the new build. Will that be a problem?

jbevain: Outside of Windows Roslyn can not fully sign assemblies. I don't know if the new msbuild tooling account for that.

Mono.Cecil.props
```diff
@@ -1,101 +0,0 @@
...
-    <Configuration Condition=" '$(Configuration)' == '' ">net_4_0_Debug</Configuration>
-    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
-    <BuildDirectory Condition=" '$(BuildDirectory)' == '' ">$(MSBuildThisFileDirectory)</BuildDirectory>
-    <FileAlignment>512</FileAlignment>
-    <SchemaVersion>2.0</SchemaVersion>
-    <OutputType>Library</OutputType>
-    <SignAssembly>true</SignAssembly>
-    <AssemblyOriginatorKeyFile>$(MSBuildThisFileDirectory)\cecil.snk</AssemblyOriginatorKeyFile>
-    <OutputType>Library</OutputType>
-    <OutputPath>$(BuildDirectory)\bin\$(Configuration)\</OutputPath>
```
sharwell: :memo: The output paths did change in this PR, but that didn't seem to be a particular problem. It's certainly easier to manage the build configurations/frameworks with the default directory layout.

jbevain: It's very likely that is going to impact users that are building from source, which is a common scenario.

jbevain:
If we take a step back, why would we do this?

We have a solution that works for previous versions of VS, and that can target .netstandard and .net desktop.

What's a compelling reason to take this change?

jbevain:
@sharwell saw from your profile you're based in Houston. I hope all is well!

NetStandard.props
sharwell: Since the Travis build passed, I'm guessing it does account for it? But just a guess.

sharwell:
> We have a solution that works for previous versions of VS, and that can target .netstandard and .net desktop.

Projects using the new project system cannot use a `ProjectReference` to reference these projects when building from source.

jbevain:
We should then also investigate if it we can add support to what's missing to support being a `ProjectReference`.

SimonCropp:
@jbevain 

re "What's a compelling reason to take this change?"

I am not sure it is compelling. but this is my experience (not all of these are applicable to the cecil repo)

 * Required for a repo to support net standard/net core.
 * approx 80% smaller csproj files
 * much easier to understand and manage the new project format
 * Easier +faster to maintain nugets by hand
 * No file locking of msbuild tasks during nuget changes
 * Less chance of merge conflicts
 * Faster restore due to shared global cache
 * Uses less disk space due to shared global cache
 * Reduced git repo bloat due to no "source packages" needing to be committed

Directory.Build.props
```diff
@@ -0,0 +1,44 @@
...
+
+  <PropertyGroup Condition="$(Configuration.StartsWith('Debug'))">
+    <DebugType>full</DebugType>
+    <DebugSymbols>true</DebugSymbols>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="$(Configuration.StartsWith('Release'))">
+    <DebugType>pdbonly</DebugType>
+    <DebugSymbols>true</DebugSymbols>
+  </PropertyGroup>
```
SimonCropp: i think the two property groups are redundant? the defaults for release and debug should suffice

Test/Mono.Cecil.Tests.csproj
```diff
@@ -1,31 +1,30 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="4.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Project Sdk="Microsoft.NET.Sdk">
+
   <PropertyGroup>
-    <ProjectGuid>{A47B1F49-A81A-43E8-BE6B-DD28AF2C4055}</ProjectGuid>
-    <RootNamespace>Mono.Cecil.Tests</RootNamespace>
-    <AssemblyName>Mono.Cecil.Tests</AssemblyName>
+    <TargetFrameworks>net40</TargetFrameworks>
```
SimonCropp: if u add multiple frameworks u can run the tests against each of those

symbols/pdb/Test/Mono.Cecil.Pdb.Tests.csproj
```diff
@@ -1,30 +1,29 @@
...
-<Project ToolsVersion="4.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Project Sdk="Microsoft.NET.Sdk">
+
   <PropertyGroup>
-    <ProjectGuid>{29300103-CB76-4A1D-B6FD-FFD91C1EC8AA}</ProjectGuid>
-    <RootNamespace>Mono.Cecil.Pdb.Tests</RootNamespace>
-    <AssemblyName>Mono.Cecil.Pdb.Tests</AssemblyName>
+    <TargetFrameworks>net40</TargetFrameworks>
+
+    <EnableDefaultItems>False</EnableDefaultItems>
```
SimonCropp: this is redundant?

Directory.Build.props
sharwell: The defaults use the new PDB format which causes a variety of errors. These conditions force the use of the old PDB format, at least when building on Windows.

Test/Mono.Cecil.Tests.csproj
sharwell: Yes, but for now I kept it the way it was being used in the past.

Directory.Build.props
jbevain: What are the variety of issues with the new PDB format?

symbols/pdb/Test/Mono.Cecil.Pdb.Tests.csproj
sharwell: No, this property can be removed (defaulted) when the project is converted to blobs. It's a bit of a challenge with the current repository because some projects are located in parent folders of other projects, which forces the use of various exclusions.

Directory.Build.props
sharwell: Very old NUnit 2 test adapter uses very old cecil, resulting in total failure to support test runs.

jbevain: :)

We could look into using NUnit 3.

jbevain:
Also I'd like to retain the AssemblyInfo.cs stuff in those files. There's a msbuild property to not generate them from the csproj.

Directory.Build.props
sharwell: > We could look into using NUnit 3.

I haven't looked into the viability of this. I do know that OpenCover doesn't support portable PDB, so I normally keep these two property groups from the start when converting a project to the new project system.

SimonCropp: ilspy also has issues. basically any tool that integrates the pdb need to be updated. so we r in a period of transition

sharwell:
> Also I'd like to retain the AssemblyInfo.cs stuff in those files. There's a msbuild property to not generate them from the csproj.

Are you opposed to generating ones that precisely match the attributes currently in use? Right now it's not quite an exact match, giving three options:

1. Use the feature in a way that exactly matches the output we have prior to this PR (specifically, disable generation of attributes such as `[AssemblyDescription("")]` which were not used prior to this PR).
2. Use the feature fully (update new attributes to have reasonable values, while exactly matching the old values for everything carried over).
3. Disable the feature, and manually duplicate the functionality via source files in the project.

I've found it beneficial to leverage the feature to the extent it supports the same values used by the old project system, and disable generation of the few attributes that deviate from this. But I can go any of the three ways.

rolfbjarne:
@jbevain we currently reference Mono.Cecil.csproj using a `ProjectReference` in [mtouch.csproj](https://github.com/xamarin/xamarin-macios/blob/f90655bd0581736b08a23bae2a8545c419050591/tools/mtouch/mtouch.csproj#L367), and although we can technically work around that (use a binary reference instead), it'll make it harder for us to debug any Cecil bugs we run into (which we do once in a while).

sharwell:
@rolfbjarne If you are (or are able to) using the new cross-platform MSBuild for building mtouch, then you could keep the project reference.

rolfbjarne:
@jbevain Yes, we can switch to msbuild, but will it continue to work in Visual Studio for Mac as well?

sharwell:
@rolfbjarne I'm not sure myself, but I left a message/link to your comment for the project system team.

KirillOsenkov:
VS for Mac supports MSBuild and the SDK-style projects. Adding @mrward to confirm.

mrward:
VS for Mac does not have good support for multi-target framework Sdk projects.

     <TargetFrameworks>net35;net40;netstandard1.3</TargetFrameworks>

It will treat the above project as though it targets .NET 3.5 only. There are also other bugs/limitations in VS Mac, such as conditions not being properly handled.

jbevain:
@sharwell 

> Projects using the new project system cannot use a ProjectReference to reference these projects when building from source. See icsharpcode/ILSpy@5975b2e.

If we're only missing those 2 properties, we can add them to netstandard.props, and then a .NET Core project can `ProjectReference` Mono.Cecil?

sharwell:
@jbevain I see two solutions for ILSpy which would work well:

1. You create a branch in this repository which follows **master** but uses the new project system.
1. We fork jbevain/cecil to icsharpcode/cecil, apply the change there, and update the ILSpy submodule to point at the fork.

The questions for you to answer are:

1. ❓ Do you want to work in the new project system or the old for primary ongoing development?
1. ❓ Do you want to make the new project system available to users who reference this repository as a submodule?

Decision matrix:

| Primary dev | Availability | Approach |
| --- | --- | --- |
| Legacy | Legacy | You do nothing, ILSpy forks and uses that |
| Legacy | Both | Merge #442 into a new branch which **master** is periodically merged into |
| New | New | Merge #442 into **master** |
| New | Both | Merge #442 into **master**, then revert #442 into a new branch that **master** is periodically merged into |

If you like the new project system offering, the last option should meet everyone's needs and eventually when the legacy projects are no longer required we can stop merging **master** into that branch.

jbevain:
I don't mind maintaining a branch for the new project system, but I'd like first to investigate what we can do to have the existing projects work for a new project system ProjectReference.

sbomer:
Some possibly useful information:

When I was bringing up the .NET core build of mono/linker, I used the new project system with ProjectReferences to cecil. Because mono/linker was still using the old project system, I added a configuration that conditionally pulls in [extra properties](https://github.com/mono/linker/blob/7c4dab5f12d9bff12653e198b0ade5c62767a1de/linker/ILLink.props) that enable it to be built from the new SDK, and removes some variables that were set in the [old project file](https://github.com/mono/linker/blob/7c4dab5f12d9bff12653e198b0ade5c62767a1de/linker/Mono.Linker.csproj). This allowed one project file to serve two purposes and be built from either project system.

The new project system had some problems with cecil's use of configurations to switch between target frameworks. This required workarounds in the consuming projects and in cecil itself.

For instance, see the description of the issue [here](https://github.com/mono/linker/blob/7c4dab5f12d9bff12653e198b0ade5c62767a1de/corebuild/integration/ILLink.Tasks/ILLink.Tasks.csproj#L160), a workaround for the restore issue in the [linker](https://github.com/mono/linker/blob/7c4dab5f12d9bff12653e198b0ade5c62767a1de/linker/Mono.Linker.csproj#L3), and a workaround for the build issue in [ILLink.Tasks](https://github.com/mono/linker/blob/7c4dab5f12d9bff12653e198b0ade5c62767a1de/corebuild/integration/ILLink.Tasks/ILLink.Tasks.csproj#L190) and in [cecil](https://github.com/jbevain/cecil/blob/a2314c895987ee801a0c3e32dac05f5399beda37/Mono.Cecil.props#L15).

I think generally, the new project system expects the use of TargetFramework to select the target, and it has good support for figuring out which TargetFramework to use when exploring dependencies specified in ProjectReferences. If we need to keep around the old project files in cecil so that we don't break existing consumers, using separate project files for the new project system might be easier than trying to make one project file support both use cases.

radekdoulik:
In Xamarin.Android (and Java.Interop) we might be able to use the new system as we don't use ProjectReference to `Mono.Cecil.csproj`.

We use the `AssemblyInfo.cs` and `Mono.Cecil.overrides` files though, to rename the assemblies and to sign them with own key.

That is what I see for now. Once we try to use the new system, there might be other issues. So I like the idea to keep the new system in the branch first or use separate project files as suggested by @sbomer

Directory.Build.props
```diff
@@ -0,0 +1,44 @@
...
+
+  <Choose>
+    <When Condition="'$(TargetFramework)' == 'net40'">
+      <PropertyGroup>
+        <DefineConstants>$(DefineConstants);NET_4_0</DefineConstants>
+      </PropertyGroup>
+    </When>
+    <When Condition="$(TargetFramework.StartsWith('netstandard'))">
+      <PropertyGroup>
+        <DefineConstants>$(DefineConstants);NET_4_0;NET_CORE</DefineConstants>
```
jbevain: We just removed NET_4_0 for netstandard in master.

jbevain:
Let's go ahead and do a branch where we'll maintain the new project system.

This branch needs to have as little changes as possible, which means not using the csproj for the AssemblyInfo stuff for instance. We also need to make sure that we provide a way for downstream users to customize the snk and the assembly names.

jbevain:
Thanks, I've landed a change to move Cecil to the new SDK style projects to target .NET Standard 2.0.

PR #444, opened 2017-09-08
Include Mono.Cecil.Tests.csproj in netstandard build

tmat:
Makes test code portable. Cross-target tests to netcoreapp1.1 (for Core CLR) and net462 when building netstandard.


tmat:
@jbevain This is a part of PR https://github.com/jbevain/cecil/pull/400, split to a separate PR as requested.

Mono.Cecil.Tests.props
```diff
@@ -2,9 +2,18 @@
 <Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
     <BuildDirectory>$(MSBuildProjectDirectory)</BuildDirectory>
+    <IsTestProject>true</IsTestProject>
   </PropertyGroup>
   <Import Project="Mono.Cecil.props" />
-  <ItemGroup>
+  <PropertyGroup Condition="'$(NetStandard)' == 'true'">
+    <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
```
jbevain: Why do we need binding redirects?

Mono.Cecil.props
```diff
@@ -36,14 +36,15 @@
...
-    <DefineConstants>$(DefineConstants);</DefineConstants>
+    <DefineConstants>$(DefineConstants);NET_3_5;</DefineConstants>
   </PropertyGroup>
   <PropertyGroup Condition=" $(Configuration.StartsWith('net_4_0')) ">
     <TargetFrameworkVersion>v4.0</TargetFrameworkVersion>
     <DefineConstants>$(DefineConstants);NET_4_0;</DefineConstants>
   </PropertyGroup>
-  <PropertyGroup Condition=" $(NetStandard) ">
-    <TargetFramework>netstandard1.3</TargetFramework>
+  <PropertyGroup Condition="'$(NetStandard)' == 'true'">
```
jbevain: Can't we use '$(NetStandard)' and '! $(NetStandard)'?

Test/Mono.Cecil.Tests/Extensions.cs
```diff
@@ -3,12 +3,57 @@
...
+		}
+
+		public static string GetCultureName (this SR.AssemblyName self)
+		{
+			return self.CultureName;
+		}
+#else
+		public static SR.Assembly LoadAssembly (MemoryStream stream)
+		{
+			return SR.Assembly.Load(stream.ToArray ());
```
jbevain: Nitpick: `Load (`

jbevain:
Thanks for contributing this!

jbevain:
Also let's move to net standard 2.0 and avoid the `GetTypeInfo()` everywhere.

Mono.Cecil.props
tmat: Most msbuild targets use ```"'$(Var)' == 'true'"``` to allow the variable to be empty. Empty string is not auto-converted to boolean false.

tmat:
Makes sense - i'll port to netstandard 2.0. I have made the change when 2.0 wasn't released yet.

tmat:
@jbevain PR updated. Now targeting netstandard 2.0. All tests pass on Core CLR.

Mono.Cecil.Tests.props
tmat: In case we cross-target tests to also run on .NET Framework.

jbevain: @tmat so we can remove this and the workaround below?

Mono.Cecil.Cil/Symbols.cs
```diff
@@ -826,7 +826,7 @@ static SR.AssemblyName GetSymbolAssemblyName (SymbolKind kind)
 
 			var suffix = GetSymbolNamespace (kind);
 
-			var cecil_name = typeof (SymbolProvider).Assembly ().GetName ();
+			var cecil_name = typeof (SymbolProvider).GetTypeInfo ().Assembly.GetName ();
```
jbevain: Can you remove all the `GetTypeInfo()` in favor of the existing API in netstd2?

Mono.Cecil.props
jbevain: Ok thanks!

Mono.Cecil.Cil/Symbols.cs
tmat: Only tests target netstandard2.0. the producut binary still targets netstandard1.3. I don't think we want to move to netstandard2.0 yet for the product binary

jbevain: Why not?

tmat: Because not everyone can move their apps to netstandard2.0. VS, for example, is on netstandard1.3 right now.

jbevain: I understand that, but do we actually have any tool that has a dependency on the netstd1.3 version of Cecil that can't move to 2.0?

tmat: I don't know any specific one. Hard to say. I think it's better to be conservative. There is not that many of these APIs to be worth preventing people from using Cecil in their apps that can't move to 2.0 yet.

tmat: I'd rather consider removing the net35 and net40 targets ...

tmat: (I mean potentially in future, not in this PR).

tmat:
@jbevain Any more feedback?

jbevain:
@tmat yes: please do switch Mono.Cecil to be built on netstandard 2.0 in this PR. Any previous usage of 0.10 was in beta and I want to release 0.10 to target netstd 2.0. After that we'll clean up a few tidbits and we should be good to go. Thanks again for contributing this!

tmat:
@jbevain Since there are currently consumers who need this change when targeting net46 (ILLinker) and also other PRs in progress that completely revamp the projects (e.g. https://github.com/jbevain/cecil/pull/442) I think it would be best to keep the netstandard1.3 for now and move to netstandard2.0 later, once the other PRs are merged. Otherwise we'll run into all sorts of issues. Either the build won't support net46 for a while or I'd need to update the configurations to cross-target and thus collide even more with https://github.com/jbevain/cecil/pull/442 than I already do. 

Sounds reasonable?

jbevain:
Sounds like a plan. I'll look into merging this tonight. Thanks!

tmat:
@jbevain Thanks!

tmat:
@jbevain Gentle reminder :)

jbevain:
Thanks for your contribution!

tmat:
@jbevain Thanks for feedback and reviews!

PR #450, opened 2017-09-30
Fix win32 resource processing.

erozenfeld:
The code assumed that win32 resources always lived in .rsrc section in the
input assembly. That's not always the case. For example, crossgen moves
win32 resources to the beginning of .text section.

The fix is to get the location of win32 resources from resources directory.
Win32 resources are still output to .rsrc section by the ImageWriter.


erozenfeld:
@jbevain PTAL

jbevain:
Thanks for contributing this Eugene!

PR #453, opened 2017-10-12
Change `BaseAssemblyResolver.SearchDirectory` to protected virtual

joelmartinez:
This allows a custom resolver to modify the behavior of searching for the assembly to load ... for example, to take the assembly's version into account rather than just loading the first one it encounters.


jbevain:
Thanks!

PR #455, opened 2017-10-18
Reorganized projects to support .Net Standard 2.0

flaviofak:
I needed support for .Net Standard 2.0, so I tried to organize the projetcs in a way that this could be naturally implemented.
The code itself is mostly untouched, but I had to make a lot of changes to the project files.
I summary, these are some of the modifications:
- All package projects were moved to directory 'src'.
- All test projects were moved to directory 'test'.
- All projects were migrated to the new .csproj format.
- NUnit was added as a nuget package in test projects. The projects can be run using Visual Studio as any MSTest or xUnit test.
- NUnit was updated to the latest version.
- Some *.props files that were added to every project were removed and a 'Directory.Build.props' file was included.
- All project information was migrated from the 'ProjectInfo.cs' and 'AssemblyFile.cs' files to the 'Directory.Build.props' file.
- I had to do some workaround to use .Net 4.6.2 in test project to run the tests for .Net Standard 2.0. .Net Core 2.0 would be natural, but many tests fail (see comments below).
- Added a project to build the NuGet package when building the project itself. There are some caveats with this approach in the comments below.

Comments:
- The class 'CodeDomProvider' doesn't exist in .Net Core, so I implemented a new 'CompilationService' using Roslyn. I think the assemblies generated using it might be incompatible with the tool 'PEVerify'. This is causing several tests to fail when the test project target is .Net Core. I don't know why exactly, so I used .Net 4.6.2 in the test projects to test .Net Standard assemblies.
- There are 2 fields in the .nuspec file that I could not replicate using project properties: <summary> and <language>. For the summary, the description is used as a fallback if not defined. But the original description was too log, so a changed the description with the summary. You can leave the full description, but it'll be cut when showing only the summary inside the NuGet page.
- There is another way to generate the package automatically using a custom .nuspec file, this covers the missing fields, but the dependecies must be added manually in the .nuspec file, so I imagined it would be better to have the dependencies generated automatically by the compiler.
- The new tooling cannot be used right now to build the project (dotnet build). This is due to the .Net 3.5 target, which is not supported by it. If the support for .Net 3.5 is dropped, the project can be compiled using the new tooling, which works in Windows, Linux and MacOS.
- The new tooling cannot be used to run the tests (dotnet test). NUnit doesn't have a test runner for it, but there is a beta version of an inplementation (dotnet-test-nunit). I didn't test it, but I don't think it'll run tests for .Net 3.5 nor 4.0 as it deppends on .Net 4.5.1.

Sorry for the massive chages, but it was mostly project organization.


jbevain:
Hello @flaviofak,

Thanks for contributing, you obviously put a lot of work into that, but I'm not going to accept this PR.

First, please read the [Open Source Contribution Etiquette](http://tirania.org/blog/archive/2010/Dec-31.html). I've worked on Cecil for many years and will continue to do so in the future. I'm not going to change its layout, I'll rather make things work with the existing layout.

Next, moving to the new project system has many implications, currently being discussed in #442.

PR #467, opened 2017-11-20
Fix reading broken MDBs emitted by mcs

TautvydasZilys:
If sequence point ends at row -1, assume that the sequence point is just for a single line. This fixes Unity case 953154.

Related PR in Unity fork: https://github.com/Unity-Technologies/cecil/pull/17

### DIFF
--- a/symbols/mdb/Mono.Cecil.Mdb/MdbReader.cs
+++ b/symbols/mdb/Mono.Cecil.Mdb/MdbReader.cs
@@ -190,7 +190,7 @@ namespace Mono.Cecil.Mdb {
 			var source = symbol_file.GetSourceFile (line.File);
 			return new SequencePoint (line.Offset, GetDocument (source)) {
 				StartLine = line.Row,
-				EndLine = line.EndRow,
+				EndLine = line.EndRow != -1 ? line.EndRow : line.Row,
 				StartColumn = line.Column,
 				EndColumn = line.EndColumn,
 			};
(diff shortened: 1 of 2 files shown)


jbevain:
@TautvydasZilys the use case that found this issue is converting MDB files to PDB right?

As far as Mono and mcs is concerned, the mdb is valid, it's just the VS debugger that doesn't like having an EndLine at -1, correct?

TautvydasZilys:
Well, I don't know what Mono considers valid or not, but EndLine being at -1 certainly doesn't sound correct. I found this when debugging assemblies on .NET that had their PDBs converted from MDBs using Cecil.

jbevain:
Looking into it, it turns out -1 is valid, and the mdb is tagged as not having line ending infos.

Ultimately, I don't think this PR is valid. It should be the responsibility of the MDB to PDB converter to make sure that the end line of the sequence point in the PDB is set to the starting line if the mdb doesn't have ending infos.

TautvydasZilys:
End lines of -1 are clearly invalid in PDBs. If it's valid in MDBs, cecil should take a stance on which behaviour it follows in its sequence points. Either behaviour is fine, as long as it is consistent. The consumer of Cecil shouldn't care whether source symbols are MDB or PDB. The reading/writing step should take care of this inconsistency.

This PR takes care of it in MDB reader. It could also be taken care of in PDB writer.

jbevain:
Right. We need to preserve the ability to roundtrip a mdb. The fix needs to go in the PDB writer.

PR #468, opened 2017-11-21
Add constructor arguments to ICustomAttribute.

rolfbjarne:
My use case is to write methods that:

* Processes constructor arguments in custom attributes.
* Accepts CustomAttribute instances
* Accepts a custom class of attributes

Adding constructor arguments to ICustomAttribute makes this possible by making
these methods take an ICustomAttribute, and have my custom class of attributes
implement ICustomAttribute.

### DIFF
--- a/Mono.Cecil/SecurityDeclaration.cs
+++ b/Mono.Cecil/SecurityDeclaration.cs
@@ -70,6 +70,14 @@ namespace Mono.Cecil {
 		{
 			this.attribute_type = attributeType;
 		}
+
+		bool ICustomAttribute.HasConstructorArguments {
+			get { return false; }
+		}
+
+		Collection<CustomAttributeArgument> ICustomAttribute.ConstructorArguments {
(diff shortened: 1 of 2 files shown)


Mono.Cecil/SecurityDeclaration.cs
```diff
@@ -70,6 +70,14 @@ public SecurityAttribute (TypeReference attributeType)
 		{
 			this.attribute_type = attributeType;
 		}
+
+		bool ICustomAttribute.HasConstructorArguments {
+			get { return false; }
+		}
+
+		Collection<CustomAttributeArgument> ICustomAttribute.ConstructorArguments {
+			get { return new Collection<CustomAttributeArgument> (); }
```
jbevain: Could you throw a NotSupportedException here?

rolfbjarne: Changed: e2176ecceba35a34c3bd5910942e22bd8f5fed14

jbevain:
Thanks!

PR #475, opened 2017-12-12
Add ILLibrary ModuleAttribute

sbomer:
This is used for R2R assemblies. This flag will be used by the linker to detect R2R assemblies, as we do here: https://github.com/mono/linker/pull/99.

We had discussed making cecil aware of R2R assemblies in https://github.com/jbevain/cecil/issues/390, but decided to put that logic in the linker. I think it's reasonable to at least add this enum value to cecil so that the linker can use it.

/cc @erozenfeld

### DIFF
--- a/Mono.Cecil/ModuleKind.cs
+++ b/Mono.Cecil/ModuleKind.cs
@@ -38,6 +38,7 @@ namespace Mono.Cecil {
 	public enum ModuleAttributes {
 		ILOnly = 1,
 		Required32Bit = 2,
+		ILLibrary = 4,
 		StrongNameSigned = 8,
 		Preferred32Bit = 0x00020000,
 	}


sbomer:
@jbevain, any thoughts?

jbevain:
@sbomer sorry this fell through the cracks of December.

No problem adding the enum value, I'm just confused about the naming. Is `IL Library` commonly used? I see you using `R2R` elsewhere? Is that `ready to run`? What does it do?

sbomer:
@jbevain, I chose ILLibrary because that's what we call the flag elsewhere, but I'm not sure why that name was chosen originally. It's defined [here](https://github.com/dotnet/coreclr/blob/5304962d18bccdfe84f08c7ffb1230c3c4d1763a/src/inc/corhdr.h#L143) in coreclr. R2R is ready to run. This flag is always set in R2R images. It may also be set in ngen images, in which case it could be necessary to check other properties of the native image to distinguish between ngen and R2R, but this is what we're using at the moment.

If you'd like more information about the flag, let me know and I can do some more digging and asking around.

jbevain:
Fair enough, thanks!

PR #477, opened 2017-12-26
Add support for Mono 4.4 API directories

Frassle:
Mono 4.4 added "reference assemblies". When resolving mscorlib we
should check for these reference assemblies at the asked for version.
Without this check mscorlib on linux will always resolve to a 4.0
assembly even if the AssemblyNameReference was for version 2.0 or 3.0.

See http://www.mono-project.com/docs/about-mono/releases/4.4.0/

Fixes opentk/opentk/issues/642

### DIFF
--- a/Mono.Cecil/BaseAssemblyResolver.cs
+++ b/Mono.Cecil/BaseAssemblyResolver.cs
@@ -281,6 +281,13 @@ namespace Mono.Cecil {
 			var file = Path.Combine (path, "mscorlib.dll");
 			if (File.Exists (file))
 				return GetAssembly (file, parameters);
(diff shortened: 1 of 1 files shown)


SimonCropp:
is it possible to add some tests?

Frassle:
I'm not sure. Maybe a test that asking for mscorlib 2 always returns mscorlib 2? But I'm not sure that's actually true, are they're still cases (even with this change) that cecil should resolve mscorlib 2 to mscorlib 4?

jbevain:
I've reformated and pushed your fix in 8577f7e064ccb44128c5d70baa084b4fefe706ad. Thanks for your contribution!

PR #479, opened 2018-01-11
Add [DebuggerDisplay] to CustomAttribute and SecurityAttribute

mrvoorhe:
When viewing CustomAttributes and SecurityAttributes collection properties on various classes in a watch window, it's tedious to find the element with the attribute type you are interested in.  Displaying the AttributeType in the debugger helps make this much easier

### DIFF
--- a/Mono.Cecil/CustomAttribute.cs
+++ b/Mono.Cecil/CustomAttribute.cs
@@ -9,7 +9,7 @@
 //
 
 using System;
-
+using System.Diagnostics;
 using Mono.Collections.Generic;
 
 namespace Mono.Cecil {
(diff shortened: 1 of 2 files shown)


Mono.Cecil/CustomAttribute.cs
```diff
@@ -66,6 +66,7 @@ public interface ICustomAttribute {
 		Collection<CustomAttributeNamedArgument> Properties { get; }
 	}
 
+	[DebuggerDisplay("{AttributeType}")]
```
jbevain: Please add a space before the `(` here and in the other.

jbevain:
Please address the nitpick and this is good to go. Thanks!

Mono.Cecil/CustomAttribute.cs
mrvoorhe: @jbevain Fixed.  Thanks!

jbevain:
Thanks!

PR #486, opened 2018-02-01
Develop/issue 485

SteveGilham:
This is as much a straw-man proposal as anything, exposing the linker version as a ushort, rather than trying to assembly a full history of possible linker versions as an enum (C++/CLI uses non-zero minor versions -- currently VS2017 v15.5.5 gives 14.12 --  and I'm not sure I can catch them all).

Watching in the debugger while the unit tests ran, we have everything from linker version 6 (a very early .net version indeed) through to 80 (VB Roslyn) coming through to the image writer, so the changes have been put through their paces.

### DIFF
--- a/Mono.Cecil.PE/Image.cs
+++ b/Mono.Cecil.PE/Image.cs
@@ -28,7 +28,7 @@ namespace Mono.Cecil.PE {
 		public string RuntimeVersion;
 		public TargetArchitecture Architecture;
 		public ModuleCharacteristics Characteristics;
-		public ushort Linker;
+		public ushort LinkerVersion;
 
 		public ImageDebugHeader DebugHeader;
 
@@ -59,7 +59,6 @@ namespace Mono.Cecil.PE {
 		public Image ()
 		{
 			counter = GetTableLength;
-			Linker = 0;
 		}
 
 		public bool HasTable (Table table)
--- a/Mono.Cecil.PE/Image.cs
+++ b/Mono.Cecil.PE/Image.cs
@@ -28,6 +28,7 @@ namespace Mono.Cecil.PE {
 		public string RuntimeVersion;
 		public TargetArchitecture Architecture;
 		public ModuleCharacteristics Characteristics;
+		public ushort Linker;
 
 		public ImageDebugHeader DebugHeader;
 
@@ -58,6 +59,7 @@ namespace Mono.Cecil.PE {
 		public Image ()
 		{
 			counter = GetTableLength;
+			Linker = 0;
 		}
 
 		public bool HasTable (Table table)
--- a/Mono.Cecil.PE/ImageReader.cs
+++ b/Mono.Cecil.PE/ImageReader.cs
@@ -81,8 +81,8 @@ namespace Mono.Cecil.PE {
 			// Characteristics		2
 			ushort characteristics = ReadUInt16 ();
 
-			ushort subsystem, dll_characteristics, linker;
-			ReadOptionalHeaders (out subsystem, out dll_characteristics, out linker);
+			ushort subsystem, dll_characteristics, linker_version;
+			ReadOptionalHeaders (out subsystem, out dll_characteristics, out linker_version);
 			ReadSections (sections);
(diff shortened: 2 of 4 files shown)


SteveGilham:
The Travis job seems inherently borked, as it didn't even start building.

Mono.Cecil/ModuleDefinition.cs
```diff
@@ -333,7 +336,12 @@ public sealed class ModuleDefinition : ModuleReference, ICustomAttributeProvider
 			set { characteristics = value; }
 		}
 
-		[Obsolete("Use FileName")]
+		public ushort Linker {
```
jbevain: I don't think we need to expose this for now, I suggest we handle this like the timestamp.

Mono.Cecil/ModuleDefinition.cs
```diff
@@ -260,6 +262,7 @@ public sealed class ModuleDefinition : ModuleReference, ICustomAttributeProvider
 		TargetArchitecture architecture;
 		ModuleAttributes attributes;
 		ModuleCharacteristics characteristics;
+		ushort linker;
```
jbevain: `internal ushort linker_version`

Mono.Cecil/ModuleDefinition.cs
```diff
@@ -240,6 +240,8 @@ public sealed class WriterParameters {
 
 	public sealed class ModuleDefinition : ModuleReference, ICustomAttributeProvider, ICustomDebugInformationProvider, IDisposable {
 
+		private static ushort defaultLinkerValue = 0; // or 8 to retain previous default behaviour
```
jbevain: I think we can remove this safely in favor of using `8` in the declaration, see below.

Mono.Cecil.PE/ImageWriter.cs
```diff
@@ -218,9 +218,8 @@ Section LastSection ()
 
 		void WriteOptionalHeaders ()
 		{
-			WriteUInt16 ((ushort) (!pe64 ? 0x10b : 0x20b));	// Magic
-			WriteByte (8);	// LMajor
-			WriteByte (0);	// LMinor
+			WriteUInt16 ((ushort) (!pe64 ? 0x10b : 0x20b)); // Magic
+			WriteUInt16 (module.Linker);
```
jbevain: `module.linker_version`

Mono.Cecil.PE/ImageReader.cs
```diff
@@ -81,15 +81,16 @@ void ReadImage ()
 			// Characteristics		2
 			ushort characteristics = ReadUInt16 ();
 
-			ushort subsystem, dll_characteristics;
-			ReadOptionalHeaders (out subsystem, out dll_characteristics);
+			ushort subsystem, dll_characteristics, linker;
```
jbevain: please standardize on `linker_version`

Mono.Cecil.PE/Image.cs
```diff
@@ -58,6 +59,7 @@ sealed class Image : IDisposable {
 		public Image ()
 		{
 			counter = GetTableLength;
+			Linker = 0;
```
jbevain: No need to initialize it here.

Mono.Cecil.PE/Image.cs
```diff
@@ -28,6 +28,7 @@ sealed class Image : IDisposable {
 		public string RuntimeVersion;
 		public TargetArchitecture Architecture;
 		public ModuleCharacteristics Characteristics;
+		public ushort Linker;
```
jbevain: `LinkerVersion` here too :)

jbevain:
Hello Steve,

Don't worry about Travis, it has its up and down.

I've added a bunch of comments, please address them, and then this will be good to go.

Thanks for your contribution!

SteveGilham:
Thanks for the review.  

With this change in place I shall be able to revert to using master and just reflectively tweak my F# binaries to look like they came from the C# Roslyn compiler.

jbevain:
Thanks for your contribution to Cecil!

PR #491, opened 2018-03-02
Fix the Document API

jbevain:


Mono.Cecil.Cil/Document.cs
```diff
@@ -62,23 +62,43 @@ public sealed class Document : DebugInformation {
...
+			get { return language_vendor.ToVendor (); }
+			set { language_vendor = value.ToGuid (); }
+		}
+
+		public Guid LanguageVendorGuid {
+			get { return language_vendor; }
+			set { language_vendor = value; }
 		}
 
 		public byte [] Hash {
```
tmat: >public byte [] Hash { [](start = 2, length = 21)

Let's also add byte[] EmbeddedSource { get; set; }

jbevain: @tmat yep. What's that about the has with the start and the length?

tmat: I'm not sure I understand your question.

jbevain: What does this mean?

tmat

tmat: Oh, that's an artifact of CodeFlow PR review tool.

jbevain: 🤣

tmat: I can see how it's very confusing in this context.

PR #492, opened 2018-03-02
null debug_info with body

EdHarvey:
debug_info is intimately related to the body.
If the body is set to null then the debug_info should also be set to null.

**Notes:**
* The debug_info can be large and so memory savings from discarding the body cannot be fully realized unless the debug_info is also discarded.
* debug_info is automatically built when the body is created. If a debug_info already exists, then an InvalidOperationException is raised from deep within the build process...


jbevain:
Makes sense. Thanks for your contribution!

PR #493, opened 2018-03-02
Expose relative size of  module metadata

EdHarvey:
Expose the raw size of a module's metadata to allow load balancing of searches over large number of assemblies. The raw size of the metadata is a reasonable indicator of the relative amount of work required to search an assembly.


jbevain:
Hi!

Thanks for the PR! Right now I don't think this is something I'd like to have as part of the public API.

Thanks again!

Issue #480, opened 2018-01-12
Better reporting when loading invalid debug symbols


spouliot:
It's hard to report an accurate description of what went wrong when bad symbols are loaded.

Right now we (XI/XM) workaround this with:
https://github.com/xamarin/xamarin-macios/pull/3203/files#diff-18af9c31adf82a237cb45f1284bb2b3eR89

which is _fragile_ (at best) and I'd like a better way. I'll propose a PR with an easy solution _soonish_ but I'm open to suggestions :)

More details in
* https://github.com/xamarin/xamarin-macios/issues/3200
* https://github.com/xamarin/xamarin-macios/pull/3079#issuecomment-352787934

jbevain:
Yeah that's pretty bad. We should probably introduce a Cecil exception for mismatched symbols. @spouliot what was the PR you wanted to suggest?

jbevain:
Fixed by #494.

PR #494, opened 2018-03-02
Provide a better exception when symbols are not found or not matching & allow skipping those exceptions

jbevain:


jbevain:
This addresses #480.

@spouliot, could your review? :)

symbols/mdb/Mono.Cecil.Mdb/MdbReader.cs
```diff
@@ -33,15 +33,7 @@ public ISymbolReader GetSymbolReader (ModuleDefinition module, Stream symbolStre
 			Mixin.CheckModule (module);
 			Mixin.CheckStream (symbolStream);
 
-			var file = MonoSymbolFile.ReadSymbolFile (symbolStream);
```
jbevain: This is done through ProcessHeader

Mono.Cecil.Cil/Symbols.cs
```diff
@@ -761,6 +761,44 @@ public interface ISymbolReaderProvider {
 		ISymbolReader GetSymbolReader (ModuleDefinition module, Stream symbolStream);
 	}
 
+#if !NET_CORE
+	[Serializable]
+#endif
+	public sealed class SymbolNotFoundException : FileNotFoundException {
```
spouliot: minor: there's an inconsistency between `Symbol` and `Symbols` in the API
e.g. `GetSymbolReader` and `ReadSymbols`
I think it's more the later (plural) but I don't care very much - still you should pick an official *winner* before adding more :)

Mono.Cecil.Cil/Symbols.cs
```diff
@@ -761,6 +761,44 @@ public interface ISymbolReaderProvider {
...
+			: base (info, context)
+		{
+		}
+#endif
+	}
+
+#if !NET_CORE
+	[Serializable]
+#endif
+	public sealed class SymbolNotMatchingException : InvalidOperationException {
```
spouliot: same: `Symbol[s]`

PR #503, opened 2018-03-22
Develop/issue 502

SteveGilham:
Based on an idiom I've used before for cases where API versioning is critical, compute a signature of the effective interface of the Mono.Cecil assembly, and break the build, with an indication that the assembly version should be stepped, if there has been a change.


SimonCropp:
From my perspective this was a problem with nunits choice of using an unstable in a non-isolated appdomain. no changes should be required by mono cecil. 

@jbevain so i think this should be closed

PR #504, opened 2018-03-23
Add a ClassLayoutTable row for structs with no instance fields

atykhyy:
Value types with no instance fields must specify a class size of 1 to be verifiable.


jbevain:
Thanks!

PR #501, opened 2018-04-01
Out-of-range rids emitted in metadata tables

atykhyy:
...if a `TypeReference` to a nested type is created with `DeclaredType` set to a `TypeDefinition`. This behavior is unexpected because `TypeDefinition` inherits from `TypeReference` and is usually usable where a `TypeReference` is required. Context: it is sometimes easier to create a `TypeReference` to a nested type that will be created later.

[MWE.zip](https://github.com/jbevain/cecil/files/1837320/MWE.zip)

### DIFF
--- a/Mono.Cecil/AssemblyWriter.cs
+++ b/Mono.Cecil/AssemblyWriter.cs
@@ -1327,8 +1327,14 @@ namespace Mono.Cecil {
 			if (type == null)
 				return MetadataToken.Zero;
 
-			if (type.IsDefinition)
-				return type.token;
+			for (var temp = type; ; temp = temp.DeclaringType)
+			{
+				if (temp.IsDefinition)
+					return type.CheckedResolve ().token;
+
+				if (temp.IsTypeSpecification () || !temp.IsNested)
+					break;
+			}
 
 			if (type.IsTypeSpecification ())
 				return GetTypeSpecToken (type);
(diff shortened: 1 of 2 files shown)


atykhyy:
The root cause of the problem was that for a nested type `MetadataBuilder.GetScopeToken` calls `GetTypeRefToken`, which expects its argument to be a `TypeReference`. If it happens to be a `TypeDefinition`, the subsequent call to `AddTypeReference` overwrites `TypeDefinition.token` with a `TypeRef` token, leading to unpredictable but usually incorrect results. Adding a check and making `AddTypeReference` throw away the `TypeRef` row and return `TypeDefinition.token` does not work because a scope token cannot be a `TypeDef` token.

atykhyy:
Despite having been filed on April 1, this is a real issue - @jbevain could you check it out?

jbevain:
Hi @atykhyy,

Sorry for the delay. I'm afraid that's not something we'll take. Although you can represent it with the Cecil object model, you can't from a metadata perspective have a nested TypeRef have its resolution scope set to a TypeDef.

If you need to create a TypeRef instead of using the nested TypeDefinition, you need to recreate a TypeReference for its declaring type too.

atykhyy:
I understand. It's awkward because `TypeDefinition` is allowed in many other contexts (base class, constraints, implemented interface, event type). This means that Cecil users must carefully check every assignment to `MemberReference.DeclaringType`, or replace the assignment with a helper method that would create the `TypeReference` if required. I don't know if it would be a good idea to put this logic into `MemberReference.DeclaringType` setter, as this could break existing code, but maybe it would make sense to add this helper method as an extension and have `AssemblyWriter` throw an exception (alerting the user early) rather than silently emitting invalid metadata? The latter doesn't seem like useful behavior to me.

PR #513, opened 2018-04-28
Add support for delay-signing with a public only keyfile

ryknuth:
Problem:
Mono.Cecil attempts to strong sign an assembly with a StrongNameKeyPair that only contains a public key. This throws an exception as a private key is required to complete the sign. However, it should be supported to place the public key into the AssemblyName without signing the assembly.

Fix:
Allow clients to pass in the public key.  Clients should determine if they have a proper StrongNameKeyPair that includes the private key.  If not, pass along the keyfile to be used as the public key.  In the case of using the public key bytes, do not attempt to sign the assembly and do not mark it as signed.

### DIFF
--- a/Mono.Cecil/AssemblyWriter.cs
+++ b/Mono.Cecil/AssemblyWriter.cs
@@ -109,7 +109,10 @@ namespace Mono.Cecil {
 				module.Attributes |= ModuleAttributes.StrongNameSigned;
 			}
 
-			if( parameters.PublicKeyBytes != null && name != null ) {
+			if (parameters.PublicKeyBytes != null && name != null) {
+				if (parameters.StrongNameKeyPair != null)
+					throw new ArgumentException ("Provide either the public key bytes or the strong name key pair, not both.");
+
 				// Delay signing with a public-only key is possible. We only need to assign the
 				// public key to the assembly name and we should not actually sign the assembly.
 				name.PublicKey = parameters.PublicKeyBytes;
--- a/Mono.Cecil/AssemblyWriter.cs
+++ b/Mono.Cecil/AssemblyWriter.cs
@@ -108,6 +108,12 @@ namespace Mono.Cecil {
 				name.PublicKey = parameters.StrongNameKeyPair.PublicKey;
 				module.Attributes |= ModuleAttributes.StrongNameSigned;
 			}
+
+			if( parameters.PublicKeyBytes != null && name != null ) {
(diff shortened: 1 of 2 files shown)


SimonCropp:
@jbevain thoughts on this one?

jbevain:
Yes yes yes. Just rename `PublicKeyBytes` to `PublicKey` and this is good to go. Thanks for the contribution!

jbevain:
Also a test would be nice.

jbevain:
Reviewing this, why not simply setting the PublicKey of the assembly name before writing?

jbevain:
@SimonCropp I don't see why this is needed and why you can't create a delay signed without having this new parameter. As far as I can tell, you can create an AssemblyDefinition with the proper flags and the proper public key, and it will be considered delay signed.

SimonCropp:
@jbevain i will defer to @ryknuth on that. since he said it was required

ryknuth:
@jbevain and @SimonCropp Isn't that what I'm doing?  I'm passing in the PublicKey to be set into the AssemblyDefinition that is used to write the assembly.  There's already the StrongNameKeyPair for signing with a private key.  If we want to allow delay signing, we should allow people to pass in the public key bytes.

jbevain:
@ryknuth yeah but you can already set the PublicKey on the assembly definition:

`assembly.Name.PublicKey = publicKey;`

Without having to add a new API to the `WriterParameters`.

ryknuth:
I see, let me paste some code here to ensure I understand.  The change would be completely in Fody.

```
        var parameters = new WriterParameters
            {
#if NET46
                StrongNameKeyPair = StrongNameKeyPair,
#endif
                WriteSymbols = debugWriterProvider != null,
                SymbolWriterProvider = debugWriterProvider,
            };

      ModuleDefinition.Assembly.Name.PublicKey = KeyFileBytes;
      ModuleDefinition.Write(AssemblyFilePath, parameters);
```

The added line is this:
`ModuleDefinition.Assembly.Name.PublicKey = KeyFileBytes;`

Am I understanding correctly?

jbevain:
I think it looks correct. You might have to turn on a flag as well on the Module:

`	module.Attributes |= ModuleAttributes.StrongNameSigned;`

ryknuth:
With delay signing, we shouldn't set StrongNameSigned if I recall correctly since it actually isn't signed.

PR #534, opened 2018-09-09
Use the 1.0 embedded PDB version

ltrzesniewski:
This sets the embedded PDB debug directory version to be consistent with what Roslyn emits and what System.Reflection.Metadata expects, as documented [here](https://github.com/dotnet/corefx/blob/master/src/System.Reflection.Metadata/specs/PE-COFF.md). 

Currently, using System.Reflection.Metadata to read embedded debug info produced by Cecil throws. The relevant check is [here](https://github.com/dotnet/corefx/blob/52121c8d0376dcb0f69570e793b6d40232ccabf9/src/System.Reflection.Metadata/src/System/Reflection/PortableExecutable/PEReader.EmbeddedPortablePdb.cs#L52-L70).


jbevain:
Thank you!

PR #538, opened 2018-09-19
Added additional `ResolutionException` overload

joelmartinez:
Allowing code to include the inner exception can add important context when there's an issue resolving ... particularly with custom resolvers.


jbevain:
Hey @joelmartinez,

Thanks for this. Can you just remove the usage of `nameof`. I'm not sure all downstream can use it.

Then it's good to go!

joelmartinez:
@jbevain you know ... I'm chastising myself now, because I completely thought at the time that I was making the change, whether I should be introducing an "unrelated" change with this patch. I've made the modification; cheers :)

jbevain:

PR #527, opened 2018-07-07
Migrate new sdk

bhugot:
I duplicated .sln and .csproj with .LegacyProject in name
I upgraded to netstandard2.0

I know there was another PR on this but as you didn't merged it it's my proposal

PR #542, opened 2018-09-26
Port Mono.Cecil to .NET Standard 2.0.

alexdrl:
### Changed the target runtime to .NET Standard 2.0.

Removed some conditional compilation ifs that were
removing the ability to strong name code.
Bring back code in BaseAssemblyResolver that builds on
.NET Standard 2.0, enabling the search of assemblies in the
GAC, allowing the code to find assemblies in the GAC, even when
referenced from a net461 DLL in .NET Framework full.

I am not sure if the .NET Standard 2.0 DLL assemblies referenced in the nuspec file needs to be updated or not.

I ran the tests in .NET Standard 2.0 and all worked.

This PR is motivated by https://github.com/dsplaisted/strongnamer/pull/31, in an effort to make it available on Linux. Sadly, this is not fully possible as of https://github.com/dotnet/corefx/issues/11225.

### DIFF
--- a/Mono.Cecil.nuspec
+++ b/Mono.Cecil.nuspec
@@ -16,7 +16,7 @@
     <dependencies>
       <group targetFramework=".NETFramework3.5" />
       <group targetFramework=".NETFramework4.0" />
-      <group targetFramework=".NETStandard1.3">
+      <group targetFramework=".NETStandard2.0">
         <dependency id="System.Collections" version="4.0.11" />
         <dependency id="System.IO.FileSystem" version="4.0.1" />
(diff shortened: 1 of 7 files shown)


alexdrl:
After creating this PR, I noticed that this PR https://github.com/jbevain/cecil/pull/527 seems to do the same, but a lot more (changing the whole SDK and files, a more complete solution). Feel free to close this PR if the other solution is better or more suitable for the Project.

jbevain:
Thank you for your contribution! We landed a different set of changes to move to the new SDK and to .NET Standard 2.0.

PR #545, opened 2018-10-03
Do not skip trailing null char in custom attribute strings

### LINKED ISSUE #544: Cecil cannot process \0 characters in string attribute value
When Cecil encounters \0 characters at the end of CA string it truncates them.

Something that in C# looks like
```c#
[A ("ABC\0")]
```

is compiled into
```
.custom instance void A::.ctor(string) = ( 01 00 04 41 42 43 00 00 00 )                      // ...ABC...
```

but roundtrip with Cecil shortens the string to
```
.custom instance void A::.ctor(string) = ( 01 00 03 41 42 43 00 00 )
```

jbevain:
Fixes #544.


marek-safar:
Thanks for the fix!

PR #547, opened 2018-10-19
Fix loading symbol assembly from GAC

FixBo:


jbevain:

PR #551, opened 2018-11-25
Fix #550: Fix flag test to correctly skip parameters

### LINKED ISSUE #550: Parameter debug info is shown as local variables
**Description**
Sometimes `ScopeDebugInformation.Variables` mistakenly lists parameters along with local variables. I bumped into it while reading info from module built with C++ CLI.

**Sample scenario**
```
// C++ CLI code. Compile into managed code (/clr flag).
void MyClass::Foo(MyStruct^ s)
{
    int a0 = ...;
    int a1 = ...;
    int a2 = ...;
    ...
}
```
`ScopeDebugInformation.Variables` generated for the above function contains local vars `a0, a1, a2` with indexes 0, 1, 2 and parameter `s` with index 1. It is impossible to determine from the info provided by Cecil which of the two items with index 1 holds the correct variable name (`a1`). The parameter info should have been filtered out.

**Root cause**
Internally, PDB stores flags for each "scope slot" (local var/parameter), the variables in the above example have `flags = 0` while the parameter has `flags = 3`. Cecil's `NativePdbReader` does not test for these flags in a correct fashion. The flags are officially documented at https://github.com/Microsoft/microsoft-pdb/blob/master/include/cvinfo.h#L3068

vgalka-sl:
Some parameters (for example in C++ CLI code) where mistakenly counted as
local variables. The flags are a bit-mask and should be tested with a mask.

Official MS documentation of the flags: https://goo.gl/RF9dXy

### DIFF
--- a/symbols/pdb/Mono.Cecil.Pdb/NativePdbReader.cs
+++ b/symbols/pdb/Mono.Cecil.Pdb/NativePdbReader.cs
@@ -183,12 +183,12 @@ namespace Mono.Cecil.Pdb {
 				parent.variables = new Collection<VariableDebugInformation> (scope.slots.Length);
 
 				foreach (PdbSlot slot in scope.slots) {
-					if (slot.flags == 1) // parameter names
+					if ((slot.flags & 1) != 0) // parameter names
 						continue;
 
 					var index = (int) slot.slot;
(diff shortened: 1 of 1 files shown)


jbevain:

PR #563, opened 2019-01-21
Typos

github-jane-doe:


jbevain:
Thank you for your contribution, but the CCI and the Mono.SymbolWriter code are downstream and making bulk changes like that are preventing clean merges, as such we generally don't apply bulk changes.

github-jane-doe:
:'(

PR #570, opened 2019-02-08
Fix off-by-one error in SignatureReader.ReadUTF8String()

akoeplinger:
If `position + length` was exactly `buffer.Length` we'd return an empty string instead of the correct value.

Unfortunately it seems to be hard to recreate this as a testcase that works on Windows.

### DIFF
--- a/Mono.Cecil/AssemblyReader.cs
+++ b/Mono.Cecil/AssemblyReader.cs
@@ -3769,7 +3769,7 @@ namespace Mono.Cecil {
 			if (length == 0)
 				return string.Empty;
 
-			if (position + length >= buffer.Length)
+			if (position + length > buffer.Length)
 				return string.Empty;
 
 			var @string = Encoding.UTF8.GetString (buffer, position, length);


jbevain:
@akoeplinger can you simply add a binary test case to Resources/assemblies built on !Windows and make a test case out of it?

akoeplinger:
@jbevain I can, but it doesn't repro with the binary built on !Windows. I haven't looked too much into why that is, the bug is pretty clear :)

jbevain:
I agree the bug is clear, but obviously I can't be trusted to not break it again :) You must have a binary somewhere that reproduces this, can't we add it to the resources?

akoeplinger:
What I meant was that I copied the binary that shows the problem on Mono/Mac to Windows and tried to make a testcase, but for some reason the `reader.image.BlobHeap.data` in [SignatureReader](https://github.com/jbevain/cecil/blob/master/Mono.Cecil/AssemblyReader.cs#L3253) has a size of 184 on mono and 132 on windows/.net framework on the exact same .dll so it doesn't reproduce.

tg73:
I have just run into this problem using 10.3. What we saw was that the value for a custom attribute property (`System.Runtime.SerializationDataMemberAttribute.Name`) "magically" became an empty string after the assembly was touched by Cecil. We only noticed because `DataContractSerializer` throws at runtime when it encounters the empty `Name` property. My issue search here did not find anything, so I started debugging myself. The blob for the attribute in question was at the end of the blob heap.

To add to the fun, the problem only shows up with assemblies touched by Cecil on our build servers (Windows Server Core 10.0.17134.0, using Build Tools for Visual Studio 15.9.9 and TeamCity), but not on development machines (Windows 10, using Visual Studio Professional 15.9.9). We use Cecil for some custom post-compilation transformations during the build process.

To me this seems like quite a serious bug which should warrant a hotfix release, as under the right circumstances it will corrupt metadata in ways which may cause incorrect but non-terminal behaviour in affected assemblies that is extremely confusing and difficult to diagnose. I was very lucky to see an exception at runtime.

xen2:
I have also ran into the issue (one custom attribute (Type) was null instead of the expected Type).
Would it be possible to release a 0.10.4 (note: fix currently not in release/0.10 branch) or 0.11 (fix already in master branch)?

jbevain:
@tg73 @xen2, yep, I'll push a `0.10.4` with the fix. Thanks!

jbevain:
@tg73 @xen2 done!

PR #578, opened 2019-04-18
Add support for emitting deterministic guids.

vargaz:


Mono.Cecil/ModuleDefinition.cs
```diff
@@ -209,6 +209,8 @@ public sealed class WriterParameters {
 			get { return key_pair; }
 			set { key_pair = value; }
 		}
+
+		public bool DeterministicGuid { get; set; }
```
jbevain: Maybe just rename to DeterministicMvid to match the public API of the module?

jbevain:
Let's just do the suggested rename and that's good to go. Thank you!

jbevain:
Thanks for the PR!

PR #594, opened 2019-06-20
Add support for custom attributes on generic constraints

jbevain:
⚠️ This is a breaking API change.


jbevain:
This is required to support https://github.com/dotnet/roslyn/issues/29997

jcouv:
Tagging @sbomer @marek-safar @agocke. 
We've just realized that supporting this scenario (attributes on type constraints) is impacting building corefx with nullability annotations because ILLinker currently doesn't understand such attributes and so drops them. 
This is not blocking Core preview 7 (corefx can just remove the one instance of such an attribute so far). 

We're trying to understand what it would take to get this change flowing into ILLinker for the next preview. 
For instance, does this require a code change from ILLinker, or would it just be a matter of referencing a new version of Cecil?

jbevain:
To reach the linker, this would need to flow to `mono/cecil` and `mono/linker` will need to react to the API breaking change, which should be minimal.

Mono.Cecil/AssemblyReader.cs
```diff
@@ -243,12 +243,20 @@ void ReadGenericParameters (IGenericParameterProvider provider)
...
 
 				ReadCustomAttributes (parameter);
 			}
 		}
 
+		void ReadGenericParameterConstraints (GenericParameter parameter)
+		{
+			var constraints = parameter.Constraints;
+
+			for (int i = 0; i < constraints.Count; i++)
```
SimonCropp: @jbevain out of curiosity, why do u prefer for over foreach?

jbevain: This codebase was started at a moment where `mcs` didn't generate great code for `foreach` on generic collections, and so used `for` instead of `foreach`. It's now used only for consistency with the rest of the code.

PR #598, opened 2019-07-19
Doxygen documentation

ratijas:
Add a simple Doxygen config in order to generate documentation from Cecil sources.

Now, I know this is not quite a 'documentation' which is totally missing, but is something browsable to begin with. At least this can give a bird's-eye view of the library structure and class hierarchy.

Build: run `$ doxygen Doxyfile` from the root of the repository.
View: open `docs/html/index.html` in your browser.
Preview on Github Pages: https://ratijas.github.io/cecil/

TODO: configure hooks / deploy pipeline to update Pages on each commit.


jbevain:
Hi @ratijas,

Thanks for prototyping this. Given that I'll end up being the one to maintain it, I'll first need to evaluate what works best for me. I'll be closing this in the meantime and re-opening it if I end up using doxygen.

Thanks again!

PR #604, opened 2019-07-31
Add I8 and UI8 fields to the VariantType enum

joshpeterson:
These fields are useful for dealing with `SAFEARRAY` marshaling. They
match the definition in the `VARENUM` in the Win32 C API.


jbevain:
Thanks!

PR #618, opened 2019-09-01
implement build independent hash

danielcrenna:
per https://groups.google.com/forum/#!topic/mono-cecil/AGq0LfBdqjo

@jbevain this is an implementation of the above, very old Google Group discussion, so things may have changed since then other than API specifics. 

Let me know if, with the addition of appropriate tests, that this could be accepted, or if there are alternatives to achieving the same result.

### DIFF
--- a/Mono.Cecil.PE/Image.cs
+++ b/Mono.Cecil.PE/Image.cs
@@ -23,6 +23,7 @@ namespace Mono.Cecil.PE {
 
 		public Disposable<Stream> Stream;
 		public string FileName;
+		public int FileChecksumPosition;
 
 		public ModuleKind Kind;
 		public string RuntimeVersion;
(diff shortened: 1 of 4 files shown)


SimonCropp:
@danielcrenna is it possible to have a test that verifies these changes?

danielcrenna:
You bet, it’s an old thread so I wanted to check if this was still relevant first — that the injection of file checksums, timestamps, and module MVID would produce non-deterministic hashes for an assembly built with otherwise identical source code.

It could also be a “Nope” to implement this in core if the aim is to encourage more practical ways (such as custom attributes).

I need this on a project that needs to avoid rebuilding packages at runtime and doesn’t have access to a build pipeline that could stamp a build hash into the assembly itself.

Will submit tests this evening.

SimonCropp:
just to be clear, i am not a maintainer, and cant speak to the relevance.

PR #622, opened 2019-10-18
Fix invalid ComputeHash from ByteByffer array computation

jeromelaban:
Addresses #770, where MVID computation in deterministic mode can be incorrect


jeromelaban:
@jbevain Is it expected that lots of tests are failing with :
```
The assembly is built by a runtime newer than the currently loaded runtime, and cannot be loaded.
```

Mono.Security.Cryptography/CryptoService.cs
```diff
@@ -146,7 +146,7 @@ static void CopyStreamChunk (Stream stream, Stream dest_stream, byte [] buffer,
 
 			using (var crypto_stream = new CryptoStream (Stream.Null, sha1, CryptoStreamMode.Write)) {
 				for (int i = 0; i < buffers.Length; i++) {
-					crypto_stream.Write (buffers [0].buffer, 0, buffers [0].length);
+					crypto_stream.Write (buffers [i].buffer, 0, buffers [i].length);
```
marek-safar: ouch!

jbevain:
@jeromelaban thanks for the PR and for finding this.

Obviously the errors are unexpected. No idea what changed since it's been working fine, I suspect a change in the GitHub actions VM environment. I'll look into that, in the meantime, this should be good to go.

PR #629, opened 2019-11-07
Remove name cache that is not invalidated when renaming types

jbevain:


Rohansi:
Any chance this would be merged still? I've been caught by this multiple times.

PR #633, opened 2019-12-06
Reduce some GC pressure that adds up in several unity workloads.

lucasmeijer:


Mono.Collections.Generic/Collection.cs
```diff
@@ -99,13 +99,21 @@ public Collection ()
 			items = Empty<T>.Array;
 		}
 
+		static class EmptyArray<T> {
```
jbevain: We already have one of those, we're just not using it. It's used on line 99: `Empty<T>.Array`.

jbevain:
We just need to address the tidbit and that's good to go. Thanks for the contribution!

Mono.Cecil/GenericInstanceType.cs
```diff
@@ -26,7 +26,7 @@ public sealed class GenericInstanceType : TypeSpecification, IGenericInstance, I
 		}
 
 		public Collection<TypeReference> GenericArguments {
-			get { return arguments ?? (arguments = new Collection<TypeReference> ()); }
+			get { return arguments ?? (arguments = new Collection<TypeReference> (4)); }
```
jbevain: Is this necessary? The new Collection will be using `Empty<TypeReference>.Array` as it starts empty which is already allocated, and it will grow to `4`, the default capacity, when needed.

Mono.Collections.Generic/Collection.cs
lucasmeijer: oops, fixed.

Mono.Cecil/GenericInstanceType.cs
lucasmeijer: you're right, reverted.

jbevain:
Thanks again!

SimonCropp:
@lucasmeijer do you have any indicative numbers on the perf improvements this results in?

lucasmeijer:
@SimonCropp no, the workload I was testing on has high varience, but on many dottrace traces Collection.Grow did show up, and the two causes were the sequencepoints and the generic instantiation arguments.    This is definitely in the small fish category.

SimonCropp:
@lucasmeijer thanks for the clarification

PR #636, opened 2019-12-16
Add index based method to ILProcessor

marek-safar:
This makes it simpler to work with it when the instruction positions
are known in advance.


jbevain:
Thank you!

PR #641, opened 2020-01-24
Improves thread safety of lazy initializations

scott-ferguson-unity:
The Interlocked.Exchange calls will ensure that the same object is
returned from each call even under a race.  That's not necessarily
required, but it should also make sure that no write re-ordering issues
occur on platforms that allow it.  And it also makes the possibility
of thread safety issues explicit.

### DIFF
--- a/Mono.Cecil.Cil/MethodBody.cs
+++ b/Mono.Cecil.Cil/MethodBody.cs
@@ -55,7 +55,7 @@ namespace Mono.Cecil.Cil {
 		public Collection<Instruction> Instructions {
 			get {
 				if (instructions == null)
-					Interlocked.CompareExchange (ref instructions, new InstructionCollection(method), null);
+					Interlocked.CompareExchange (ref instructions, new InstructionCollection (method), null);
 
 				return instructions;
 			}
@@ -79,9 +79,9 @@ namespace Mono.Cecil.Cil {
 		}
 
 		public Collection<VariableDefinition> Variables {
-			get	{
+			get {
 				if (variables == null)
-					Interlocked.CompareExchange (ref variables, new VariableDefinitionCollection(), null);
+					Interlocked.CompareExchange (ref variables, new VariableDefinitionCollection (), null);
 
 				return variables;
 			}
--- a/Mono.Cecil.Cil/MethodBody.cs
+++ b/Mono.Cecil.Cil/MethodBody.cs
@@ -53,7 +53,12 @@ namespace Mono.Cecil.Cil {
 		}
 
 		public Collection<Instruction> Instructions {
-			get { return instructions ?? (instructions = new InstructionCollection (method)); }
+			get {
+				if (instructions == null)
+					Interlocked.CompareExchange (ref instructions, new InstructionCollection(method), null);
+
+				return instructions;
+			}
 		}
 
 		public bool HasExceptionHandlers {
@@ -61,7 +66,12 @@ namespace Mono.Cecil.Cil {
 		}
 
 		public Collection<ExceptionHandler> ExceptionHandlers {
-			get { return exceptions ?? (exceptions = new Collection<ExceptionHandler> ()); }
+			get {
+				if (exceptions == null)
+					Interlocked.CompareExchange (ref exceptions, new Collection<ExceptionHandler> (), null);
+
+				return exceptions;
+			}
 		}
 
 		public bool HasVariables {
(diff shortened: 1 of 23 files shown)


Mono.Cecil.Cil/MethodBody.cs
```diff
@@ -53,23 +53,38 @@ public sealed class MethodBody {
 		}
 
 		public Collection<Instruction> Instructions {
-			get { return instructions ?? (instructions = new InstructionCollection (method)); }
+			get {
+				if (instructions == null)
+					Interlocked.CompareExchange (ref instructions, new InstructionCollection(method), null);
```
jbevain: Nit pick: please add a space before the opening parentheses in `new InstructionCollection (` to match the overall codestyle.

Mono.Cecil.Cil/MethodBody.cs
```diff
@@ -53,23 +53,38 @@ public sealed class MethodBody {
...
 
 		public bool HasVariables {
 			get { return !variables.IsNullOrEmpty (); }
 		}
 
 		public Collection<VariableDefinition> Variables {
-			get { return variables ?? (variables = new VariableDefinitionCollection ()); }
+			get	{
+				if (variables == null)
+					Interlocked.CompareExchange (ref variables, new VariableDefinitionCollection(), null);
```
jbevain: nit: same.

Mono.Cecil/SecurityDeclaration.cs
```diff
@@ -144,16 +158,22 @@ public SecurityDeclaration (SecurityAction action, byte [] blob)
...
 				return;
 
-			module.Read (this, (declaration, reader) => reader.ReadSecurityDeclarationSignature (declaration));
-			resolved = true;
+			lock (module.SyncRoot) {
+
+				if (resolved)
+					return;
+
+                module.Read (this, (declaration, reader) => reader.ReadSecurityDeclarationSignature (declaration));
```
jbevain: Nit: indent looks wrong (and uses spaces instead of tabs)

Mono.Collections.Generic/ReadOnlyCollection.cs
```diff
@@ -19,7 +20,14 @@ public sealed class ReadOnlyCollection<T> : Collection<T>, ICollection<T>, IList
 		static ReadOnlyCollection<T> empty;
 
 		public static ReadOnlyCollection<T> Empty {
-			get { return empty ?? (empty = new ReadOnlyCollection<T> ()); }
+			get
+			{
```
jbevain: Nit: should be on the get line.

jbevain:
Also, I'm not quite sure what is up with the failing test on Windows. That might be a by-product of the change, even though it looks unrelated to the PR functionality.

Mono.Cecil.Cil/MethodBody.cs
scott-ferguson-unity: updated

Mono.Cecil.Cil/MethodBody.cs
scott-ferguson-unity: updated

Mono.Cecil/SecurityDeclaration.cs
scott-ferguson-unity: updated

Mono.Collections.Generic/ReadOnlyCollection.cs
scott-ferguson-unity: updated

scott-ferguson-unity:
The failing tests on Windows were my fault - `AssemblyNameReference.public_key_token` would sometimes be `Array<byte>.Empty` instead of null.  I updated so the initialized value is `null`.

I didn't think that breaks anything and keeps the pattern the same as the other cases.

jbevain:
@scott-ferguson-unity thanks for getting this green. This is good to merge on my end. Do you want to add anything before I merge?

scott-ferguson-unity:
No, I don't believe anything else needs to be added.

jbevain:
Thanks for the PR!

teo-tsirpanis:
@scott-ferguson-unity , there is actually a [dedicated `Lazy` class](https://docs.microsoft.com/en-us/dotnet/api/system.lazy-1) for that which is thread-safe. This PR brought unnecessary code duplication.

ltrzesniewski:
@teo-tsirpanis using `Lazy<T>` would cause increased GC pressure, which I guess would quickly become a problem given these are hot code paths.

teo-tsirpanis:
[`LazyInitializer.EnsureInitialized`](https://docs.microsoft.com/en-us/dotnet/api/system.threading.lazyinitializer.ensureinitialized)

ltrzesniewski:
That one is *much* better, but I think it would need to be profiled (it uses either reflection or a lambda, depending on the overload you choose).

PR #644, opened 2020-01-30
Fix locally scoped constant array

SimonCropp:
@jbevain seems to me an array should be treated the same as a object type for this context?


SimonCropp:
put in a diff PR since i am confident the fail test verifies the scenario. but am uncertain about this as a fix

Test/Mono.Cecil.Tests/ImageReadTests.cs
```diff
@@ -207,6 +207,14 @@ public void Net471TargetingAssembly ()
 			});
 		}
 
+		[Test]
+		public void LocallyScopedConstantArray ()
+		{
+			TestModule ("LocallyScopedConstantArray.dll", module => {
+				Assert.IsTrue (module.HasDebugHeader);
```
jbevain: We'd need to assert that the constant is indeed null in there.

jbevain:
Thanks for the PR. The fix is good!

Test/Mono.Cecil.Tests/ImageReadTests.cs
SimonCropp: done

jbevain:
Thanks for the PR!

PR #645, opened 2020-01-31
Clean-up some old code et. al.

teo-tsirpanis:
This PR:

* Redirects the functionality of the `Mono.Collections.Generic.Collection` class to `System.Collections.ObjectModel.Collection`. Cecil's collection type was created ten years ago and duplicates lots of logic that is already implemented in battle-tested classes like `System.Collections.Generic.List`.

* Removes `Mono.Collections.Generic.ReadOnlyCollection`. It was nowhere used, and .NET provides a similar class in the core libraries since .NET Framework 2.0. (I acknowledge this being a breaking change, albeit a minor one.

* Replaces the deprecated `licenseUrl` element with SPDX license expressions.

* Removes some unused (private) functions.

* Updates the NuGet packages to their latest versions.


jbevain:
Thank you for your interest in Cecil.

At this point however I'm not going to take this PR: I don't take broad clean-up PRs that touch a bit of everything.

I also would not take the main change that replaces the Collection implementation.

teo-tsirpanis:
I understand that the broad scope of this PR is intimidating.

Can we at least agree that Cecil's codebase is in need of some polishing?

PR #648, opened 2020-02-12
Changing PdbReaderProvider to accept create EmbeddedPDB from stream.

thaystg:
Changing GetSymbolReader from PdbReaderProvider, when we create a Module from stream  we don't have the filename, and it's not necessary to create an EmbeddedPortablePdbReaderProvider, so I've moved the check to after the creation of EmbeddedPortablePdbReaderProvider.

### DIFF
--- a/Test/Mono.Cecil.Tests/PortablePdbTests.cs
+++ b/Test/Mono.Cecil.Tests/PortablePdbTests.cs
@@ -5,6 +5,7 @@ using System.Text;
 using NUnit.Framework;
 
 using Mono.Cecil.Cil;
+using Mono.Cecil.Pdb;
 using Mono.Cecil.PE;
 
 namespace Mono.Cecil.Tests {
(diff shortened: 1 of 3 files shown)


Test/Mono.Cecil.Tests/PortablePdbTests.cs
```diff
@@ -368,6 +369,28 @@ public void EmbeddedCompressedPortablePdb ()
 			}, symbolReaderProvider: typeof (EmbeddedPortablePdbReaderProvider), symbolWriterProvider: typeof (EmbeddedPortablePdbWriterProvider));
 		}
 
+		[Test]
+		public void EmbeddedCompressedPortablePdbFromStream ()
+		{
+			TestBinary("EmbeddedCompressedPdbTarget.exe", module => {
```
jbevain: I think we can make this simpler and not create infrastructure for reading modules as in memory. You can undo all the changes in BaseTestFixture, and in this test you can use:

`var bytes = File.ReadAllBytes (GetAssemblyResourcePath ("EmbeddedCompressedPdbTarget.exe"))`

Create the module from that and run the test.

thaystg: Done!

jbevain:
Thank you!

PR #650, opened 2020-02-15
add strong key pair for NET Standard

jetersen:
Hi @SimonCropp and @jbevain I followed up on the suggestion in #548

Not sure if my 54a8987 is helpful, I can easily remove it if you feel like it's a bad addition.

For our use case it meant we could remove `using System.Reflection` and avoid the same 
```
#if (NETSTANDARD)
using StrongNameKeyPair=Mono.Cecil.StrongNameKeyPair;
#else
using StrongNameKeyPair=System.Reflection.StrongNameKeyPair;
#endif
```

In all our files that used StrongNameKeyPair.

### DIFF
--- a/Mono.Cecil/ModuleDefinition.cs
+++ b/Mono.Cecil/ModuleDefinition.cs
@@ -13,6 +13,11 @@ using System.Collections.Generic;
 using System.IO;
 using System.Threading;
 using SR = System.Reflection;
+#if (NETSTANDARD)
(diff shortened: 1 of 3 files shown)


jbevain:
Hey!

Thanks for your contribution! Unfortunately I don't think I'll take this one, I'm not sure I want to add that to the public API and support it and have people take a dependency on Cecil for that feature.

It should be very easy to take the code needed out and extract that into a separate project.

Thanks again!

PR #662, opened 2020-04-22
Add Clear to ILProcessor

MichalPetryka:
Clear can be useful when you need to rewrite a method from scratch.

### DIFF
--- a/Mono.Cecil.Cil/ILProcessor.cs
+++ b/Mono.Cecil.Cil/ILProcessor.cs
@@ -283,5 +283,10 @@ namespace Mono.Cecil.Cil {
 
 			instructions.RemoveAt (index);
 		}
+
+		public void Clear ()
+		{
+			instructions.Clear ();
+		}
 	}
 }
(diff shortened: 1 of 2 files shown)


jbevain:
Thanks!

ltrzesniewski:
Hmmm.... If you want to rewrite a method from scratch, there's a whole lot of other stuff you need to remove. I'm doing [the following in a project of mine](https://github.com/ltrzesniewski/InlineIL.Fody/blob/2b63811da7cf22473a174989f98f225a3083c136/src/InlineIL.Fody/ModuleWeaver.cs#L63-L71):

```C#
method.Body.Instructions.Clear();
method.Body.Variables.Clear();
method.Body.ExceptionHandlers.Clear();

method.DebugInformation.SequencePoints.Clear();
method.DebugInformation.Scope = null;
method.DebugInformation.StateMachineKickOffMethod = null;

method.CustomDebugInformations.Clear();
```

On the other hand, the `ILProcessor` only deals with the instruction list today, so I suppose making it clear a bunch of other stuff would not be appropriate.

jbevain:
@ltrzesniewski yeah I think that's a different subject. At that point you might be better off creating a new method body.

PR #663, opened 2020-04-27
Add a way to read PE images loaded in memory.

KirillOsenkov:
It is useful when inspecting modules loaded into a live process or reading process dump files.

On disk RVAs must be mapped to stream contents using section header mapping information. In memory RVAs can be used as direct offsets into the stream.

### DIFF
--- a/Mono.Cecil.PE/Image.cs
+++ b/Mono.Cecil.PE/Image.cs
@@ -23,6 +23,7 @@ namespace Mono.Cecil.PE {
 
 		public Disposable<Stream> Stream;
 		public string FileName;
+		public bool ImageLayoutInMemory;
 
 		public ModuleKind Kind;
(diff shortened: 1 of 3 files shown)


KirillOsenkov:
You could read modules from memory or dump files using this ReadVirtualStream.cs:

https://gist.github.com/KirillOsenkov/1b12d8cca60e6dc1c13f6859451a2aab

KirillOsenkov:
I took the implementation from the DumpModules project by @ljcollins25.
It is used to save all modules from a .dmp file using ClrMD:
https://github.com/Ref12/DumpModules/blob/master/src/Ref12.Diagnostics.Extraction/ModuleDumper.cs

gdynamics:
This would be a very useful feature.

PR #666, opened 2020-05-14
Removing the assumption that StringHeap is always present in the metadata

### LINKED ISSUE #665: [Q] Is it true that StringHeap must always be present in a portable PDB metadata?
In the following section of the code, StringHeap is treated as it will always be present in the metadata, while for Guid and Blob heaps that is not the case:

https://github.com/jbevain/cecil/blob/f74703c36e9a7626ef435399c45d27a84b246864/Mono.Cecil.PE/ImageReader.cs#L480#L486

I am working on a tool for creating portable PDB metadata (and verifying it with Cecil library), and in some cases when StringHeap is not present in the portable PDB metadata, Cecil library throws null exception at:
https://github.com/jbevain/cecil/blob/f74703c36e9a7626ef435399c45d27a84b246864/Mono.Cecil.PE/ImageReader.cs#L484 

Does StringHeap have to be present in portable PDB metadata in all cases? 
And if so, is this requirement documented anywhere?

Thank you

ivanpovazan:
Fixes #665

### DIFF
--- a/Mono.Cecil.PE/ImageReader.cs
+++ b/Mono.Cecil.PE/ImageReader.cs
@@ -481,7 +481,7 @@ namespace Mono.Cecil.PE {
 		{
 			uint offset = (uint) BaseStream.Position - table_heap_offset - image.MetadataSection.PointerToRawData; // header
 
-			int stridx_size = image.StringHeap.IndexSize;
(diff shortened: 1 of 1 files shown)


jbevain:
Thanks for your contribution!

PR #676, opened 2020-06-20
Automatically update local variable debug info

### LINKED ISSUE #674: Removing local variable from method body does not remove it from debug info
Editing method body by removing one of the local variables does not update the attached debug information by removing debug info for the local variable. When saved the PDB still has the local variable information in it and some tools get confused (for example PDB2PDB writes out a warning).

I think the correct way to handle this would be to add logic into  [`VariableDefinitionCollection.OnRemove`](https://github.com/jbevain/cecil/blob/c1a3c199c31b76e3064316a5f24e176ca780b0af/Mono.Cecil.Cil/MethodBody.cs#L164) to also remove the variable from the debug info (if it's there) by its index.

I don't think there's anything to do in `OnAdd` as adding `VariableDefinition` doesn't carry any information which should be preserved in the debug info (like the name of the variable). So not adding a debug info for the new variable seems to be OK.

Originally found in https://github.com/mono/linker/issues/1260.

I plan to work on the PR for this unless somebody wants to pick it up first.

vitek-karas:
When manipulating the local variables collection on a method body, automatically update the debug info for all affected local variables.

When removing a local variable, also remove the debug info for that local variable from the scopes.

When removing or inserting update the indices of local variable debug infos which are afected by the action. Note the local variable debug info either holds just a backpointer to the local variable in which case it doesn't store the index at all (and thus nothing to do), or it stores the index explicitly in which case it needs to be updated.

Added tests for both insert and remove cases.

Fixes #674 

/cc @marek-safar

### DIFF
--- a/Mono.Cecil.Cil/MethodBody.cs
+++ b/Mono.Cecil.Cil/MethodBody.cs
@@ -81,7 +81,7 @@ namespace Mono.Cecil.Cil {
 		public Collection<VariableDefinition> Variables {
 			get {
 				if (variables == null)
-					Interlocked.CompareExchange (ref variables, new VariableDefinitionCollection (), null);
+					Interlocked.CompareExchange (ref variables, new VariableDefinitionCollection (this.method), null);
 
 				return variables;
 			}
@@ -134,13 +134,17 @@ namespace Mono.Cecil.Cil {
 
 	sealed class VariableDefinitionCollection : Collection<VariableDefinition> {
 
-		internal VariableDefinitionCollection ()
+		readonly MethodDefinition method;
+
+		internal VariableDefinitionCollection (MethodDefinition method)
 		{
+			this.method = method;
 		}
 
-		internal VariableDefinitionCollection (int capacity)
+		internal VariableDefinitionCollection (MethodDefinition method, int capacity)
 			: base (capacity)
 		{
+			this.method = method;
 		}
 
 		protected override void OnAdd (VariableDefinition item, int index)
@@ -151,9 +155,7 @@ namespace Mono.Cecil.Cil {
 		protected override void OnInsert (VariableDefinition item, int index)
 		{
 			item.index = index;
-
-			for (int i = index; i < size; i++)
-				items [i].index = i + 1;
+			UpdateVariableIndices (index, 1);
 		}
 
 		protected override void OnSet (VariableDefinition item, int index)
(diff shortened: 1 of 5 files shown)


Mono.Cecil.Cil/MethodBody.cs
```diff
@@ -163,10 +165,35 @@ protected override void OnSet (VariableDefinition item, int index)
...
+				if (scope.HasVariables) {
+					var variables = scope.Variables;
+					for (int i = 0; i < variables.Count; i++) {
+						if (variables [i].Index == indexToRemove) {
+							variables.RemoveAt (i);
+						}
+					}
+
+					foreach (var variable in variables) {
+						if (variable.index.index.HasValue && variable.index.index.Value >= startIndex) {
```
jbevain, line 191: Because this looks a bit odd, I think we want to comment as to why we're doing this. In that case my understanding is that you want to update the indexes of variables that have not been resolved into a VariableDefinition, while a VariableIndex can represent both the resolved and not resolved case.

jbevain:
Thanks, this looks good!

jbevain:
Looking at this PR and the other one, one thing that would be neat is to add a internal property to InstructionOffset and VariableIndex, like `bool IsResolved`. This way you can have internal logic for the non resolved case, which you have in both PR, and access the relevant data without having to poke at the internals. That would avoid the `.index.index` access which does look weird.

Mono.Cecil.Cil/MethodBody.cs
vitek-karas: Hopefully the new version is more readable (and I also added a comment) - and it now also handles removal of resolved variable debug infos.

vitek-karas:
I ended up adding two new internal properties (to both `VariableIndex` and `InstructionOffset`):
* `IsResolved` per your suggestion
* `ResolvedVariable`/`ResolvedInstruction` as the code still needs to be able to access the resolved reference to check for removal of the target.

vitek-karas:
@jbevain could you please let me know if the latest state of this PR and the friend in https://github.com/jbevain/cecil/pull/677 is such that it will get merged sometime soon-ish? I'm trying to decide if we need to port it over to mono/cecil and use the fork there to ship the linker for .NET 5, or if we should keep using the cecil from this repo. (Without these fixes we are having trouble using the linker in the dotnet/runtime repo to build the libraries for .NET 5)

Mono.Cecil.Cil/MethodBody.cs
```diff
@@ -163,10 +165,46 @@ protected override void OnSet (VariableDefinition item, int index)
...
+		{
+			for (int i = startIndex; i < size; i++)
+				items [i].index = i + offset;
+
+			var debug_info = method == null ? null : method.debug_info;
+			if (debug_info == null || debug_info.Scope == null)
+				return;
+
+			foreach (var scope in debug_info.GetScopes ()) {
+				if (scope.HasVariables) {
```
jbevain, line 182: Please transform this condition into an early continue if there's no variable to reduce nesting.

jbevain:
@vitek-karas yes, please just address the nitpick and this is good to go. Thanks for the contribution!

jbevain:
Thank you!

PR #677, opened 2020-06-23
Auto update local scopes in debug info when manipulating instructions

### LINKED ISSUE #675: Removing instruction from method body can corrupt local scopes debug info
Removing an instruction from method body currently doesn't update the attached debug information's local scope ranges. This can lead to incorrect or even corrupt debug information.
An example is removing instructions from the end of the method eventually removing enough so that the last local scope in the method should be removed. When such method is written the associated PDB still has the last local scope in it, but now its start offset points after the method body end. Depending on the reader this can lead to either confusion or plain crashes.

For example the PDB2PDB tool (which uses System.Reflection.Metadata reader for the portable PDB) will either fail to read or fail to process such PDB.

 [`InstructionCollection.OnRemove`](https://github.com/jbevain/cecil/blob/c1a3c199c31b76e3064316a5f24e176ca780b0af/Mono.Cecil.Cil/MethodBody.cs#L232) already contains logic to update sequence points in the debug info accordingly. This can be updated to handle local scopes as well.

More complex problem is how to handle `OnAdd` and `OnInsert`. If the instruction is being inserted on a local scope boundary it's unclear to which scope it should fall into. The implemented behavior should works such that `ILProcessor.Replace` which internally inserts after and removes ends up updating the local scopes correctly (so that the replaced instruction ends up in the same local scope and the start/end of local scopes don't move around).

Originally found in https://github.com/mono/linker/issues/1267

I plan to work on a PR for this unless somebody wants to pick it up first.

vitek-karas:
Before this change Cecil would not update local scopes when manipulating method bodies. This can lead to corrupted debug info - which when written to a PDB results in corrupted PDB.

Cecil can store local scopes in two ways:
* Using IL offset numbers - this is how pretty much all of the symbol readers populate the OMs.
* Using references to instructions.

If the local scopes use offset values this change will update the local scopes for all insert and remove operations on the method body. The behaviors for insert is basically "insert after" in that the new instruction is added to the scopes of the previous instruction.

If the local scopes are using instructions directly the change only removes any references to instructions which are being removed from the method body (and replaces them with the instruction offset value).

To be able to tell the difference between these cases the `instruction` field has been made internal in the `InstructionOffset` structure.

Fixes #675.

### DIFF
--- a/Mono.Cecil.Cil/MethodBody.cs
+++ b/Mono.Cecil.Cil/MethodBody.cs
@@ -293,12 +293,12 @@ namespace Mono.Cecil.Cil {
 			//   - if the scope uses instruction reference to an instruction being removed - update it to remove the reference to the removed instruction
 			//   - if the scope uses any other instruction there's really no need to update it.
 			if ((!scope.Start.IsResolved && scope.Start.Offset >= startFromOffset) || 
-				(instructionRemoved != null && scope.Start.instruction == instructionRemoved))
+				(instructionRemoved != null && scope.Start.ResolvedInstruction == instructionRemoved))
 				scope.Start = new InstructionOffset (scope.Start.Offset + offset);
 
 			if (!scope.End.IsEndOfMethod && 
 				((!scope.End.IsResolved && scope.End.Offset >= startFromOffset) ||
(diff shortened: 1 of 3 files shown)


jbevain:
Thank you!

Issue #680, opened 2020-07-16
MdbWriter doesn't write the correct MVID when using deterministic MVID


vitek-karas:
MdbWriter caches the MVID to write upon construction of the writer from the module's Mvid property. But in case of deterministic MVID the actual MVID is computed after the MdbWriter is constructed. So the symbols end up using the wrong (Typically all zeroes) MVID.

I looked at how PortablePdb does this, and the writer in this case holds on to the ModuleDefinition and reads its Mvid property only when actually writing the symbols (which is after it was populated with the right value).

jbevain:
@vitek-karas good catch, could up send your PR upstream? Thanks!

vitek-karas:
@jbevain ported to upstream in https://github.com/jbevain/cecil/pull/681

PR #681, opened 2020-07-20
Fix MdbWriter to produce the correct MVID in case deterministic output is required

vitek-karas:
Fixes https://github.com/jbevain/cecil/issues/680.

The MdbWriter must not cache the MVID upon writer construction, but it needs to get it only after it was computed - so basically only when it's about to actually write.


jbevain:
Thank you!

PR #686, opened 2020-08-12
Fix ReadSymbols in a Module that is already created

thaystg:
When I try to use ReadSymbols in a Module that is already created, for example:

```
var symbolReader = portablePdbReaderProvider.GetSymbolReader(asm.image, stream);
asm.image.ReadSymbols(symbolReader);
```

method.debug_info has a list, but it's empty, so it wasn't entering in the if, but it should, maybe we should change the if to method.debug_info == null || method.debug_info.count = 0.

Relates to this PR: https://github.com/dotnet/runtime/pull/40690


jbevain:
Hi @thaystg, good to see you here.

Could you add a unit test that reproduces your scenario?

thaystg:
> Hi @thaystg, good to see you here.
> 
> Could you add a unit test that reproduces your scenario?

Done!!! :)

thaystg:
@jbevain could you please merge and generate a new nuget version? We need this for wasm debug.

jbevain:
Thanks!

jbevain:
@thaystg the fix shipped in 0.11.3. Thanks!

Issue #689, opened 2020-09-02
Cannot read public key of a resource dll


vasily-kirichenko:
A dll (something called "WinZip Google Drive Service", https://drive.google.com/file/d/1ymJ0Vl4JbRhy4EHY26NsrQGVhVT9JMrQ/view?usp=sharing) has valid strong name:
```
C:\Program Files (x86)\Microsoft Visual Studio\2019\Community>sn -Tp 76C329AD1DB92F3197ECF590104232909783B5E4B8D5983E2766D923EC7FF7CC

Microsoft (R) .NET Framework Strong Name Utility  Version 4.0.30319.0
Copyright (c) Microsoft Corporation.  All rights reserved.

Public key (hash algorithm: sha1):
0024000004800000940000000602000000240000525341310004000001000100a90c1fac649c95
d5c8e99e33aa6f28487c1bfc482b3fdf937fb7133b18b758cb33d0204b3fe61432c3a7acbabed8
1806dfec23376d5b338d98d122a3992c6013ac565d9a64d4b3487a5343c754dbdbd9e94b14ff77
f762862b6c8bbc50124313eb0b90b82867084cf0bcc92c8823562af1cef9c1469703f7752c4f7e
0a7a39c3

Public key token is 3e104464bfd77002
```
`Mono.Cecil.AssemblyDefinition.ReadAssembly` returns empty public key:

![image](https://user-images.githubusercontent.com/873919/91974408-68bd5480-ed26-11ea-96a1-cb3cc74452c7.png)

jbevain:
@vasily-kirichenko the assembly is obfuscated and designed to break code like Cecil. I'll see if there's something quick I can do, but if not, don't keep your hopes high :)

jbevain:
Afraid not :)

PR #690, opened 2020-09-02
add failing tests for #689

SimonCropp:


SimonCropp:
@jbevain this one is weird. if i set a breakpoint in the set of AssemblyNameReference.HashAlgorithm, it is hit 3 time. the first two with sha1, the last time with `0x0182003b`. so not sure if it is a corrupted assembly or a bug in cecil. thoughts?

Issue #694, opened 2020-10-02
Sequence points can no longer be cleared when using deferred reading mode


ltrzesniewski:
This is a regression introduced by #686. When using the deferred reading mode, if you clear the sequence points of a method, those will be read again from the source file on writing.

Here's a repro:

```C#
using System;
using System.Linq;
using Mono.Cecil;

namespace ConsoleApp
{
    internal static class Program
    {
        public static void Main()
        {
            using var assembly = AssemblyDefinition.ReadAssembly(typeof(Program).Assembly.Location, new ReaderParameters { ReadSymbols = true });
            var method = assembly.MainModule.GetType(typeof(Program).FullName).Methods.Single(m => m.Name == nameof(SomeMethod));

            Console.WriteLine($"Sequence points A: {method.DebugInformation.SequencePoints.Count}");

            method.DebugInformation.SequencePoints.Clear();

            Console.WriteLine($"Sequence points B: {method.DebugInformation.SequencePoints.Count}");

            assembly.Write("Processed.dll", new WriterParameters { WriteSymbols = true });

            Console.WriteLine($"Sequence points C: {method.DebugInformation.SequencePoints.Count}");
        }

        private static void SomeMethod()
        {
            Console.WriteLine("Hello!");
        }
    }
}
```

This prints the following on v0.11.3:

```
Sequence points A: 3
Sequence points B: 0
Sequence points C: 3
```

It prints the expected result on v0.11.2, or if you set `ReadingMode = ReadingMode.Immediate` in `ReaderParameters`:

```
Sequence points A: 3
Sequence points B: 0
Sequence points C: 0
```

I'm not entirely sure what the proper fix would be here. Maybe an additional flag is needed to tell if debug information has been read, so the status does not depend on sequence points.

/cc @thaystg FYI

jbevain:
@ltrzesniewski good catch. Could you submit a failing test? Thank you!

thaystg:
If you need to revert it's okay for me I did a workaround on wasm debugger side.

jbevain:
Cool, let's do that then.

jbevain:
Fixed in master. Thanks!

PR #695, opened 2020-10-02
Add failing test for #694

ltrzesniewski:
This adds a failing test case for #694.

I've checked that removing `|| !method.debug_info.HasSequencePoints` in the `ReadMethodsSymbols` method makes the test pass.


jbevain:
Thanks @ltrzesniewski.

jbevain:
Thank you @ltrzesniewski!

Issue #697, opened 2020-10-14
Regression in 0.11.3 -- NRE from ILProcessor.InsertBefore()


SteveGilham:
```
System.NullReferenceException
  HResult=0x80004003
  Message=Object reference not set to an instance of an object.
  Source=Mono.Cecil
  StackTrace:
   at Mono.Cecil.Cil.InstructionCollection.ResolveInstructionOffset(InstructionOffset inputOffset, InstructionOffsetCache& cache) in C:\sources\cecil\Mono.Cecil.Cil\MethodBody.cs:line 412
   at Mono.Cecil.Cil.InstructionCollection.UpdateLocalScope(ScopeDebugInformation scope, Instruction removedInstruction, Instruction existingInstruction, InstructionOffsetCache& cache) in C:\sources\cecil\Mono.Cecil.Cil\MethodBody.cs:line 345
   at Mono.Cecil.Cil.InstructionCollection.UpdateLocalScope(ScopeDebugInformation scope, Instruction removedInstruction, Instruction existingInstruction, InstructionOffsetCache& cache) in C:\sources\cecil\Mono.Cecil.Cil\MethodBody.cs:line 356
   at Mono.Cecil.Cil.InstructionCollection.UpdateLocalScopes(Instruction removedInstruction, Instruction existingInstruction) in C:\sources\cecil\Mono.Cecil.Cil\MethodBody.cs:line 334
   at Mono.Collections.Generic.Collection`1.Insert(Int32 index, T item) in C:\sources\cecil\Mono.Collections.Generic\Collection.cs:line 149
```

Simple repro attached

[Cecil.11.3.issue.zip](https://github.com/jbevain/cecil/files/5377742/Cecil.11.3.issue.zip)

SteveGilham:
The null reference is the 7th of the 6 instructions in the method body; the initial access to the instructions through the `Collection<Instruction>` type with size less than capacity decays to a direct access to the internal `Instruction[]` which has length equal to the capacity.

FWIW, the assembly was built with an old Mono compiler, so advertises a code size of 14, hence the capacity of 7, for 6 instructions.

jbevain:
Fixed in #698.

PR #698, opened 2020-10-14
A minimum tactical fix for issue #697

SteveGilham:
Terminate the iteration if it steps on to a null `Instruction` to allow for uninitialized trailing cells in the case of size less than capacity.


jbevain:
Hello @SteveGilham,

Thank for the PR! Could add a unit test to the test suite from your repro before I merge? Thanks!

Jb

SteveGilham:
The test does a little more manipulation than the plain "it blows up" repro to

1. match the full use case in which the issue arose
2. get to an end state that is simple to assert

I don't know if there's a more elegant way in the infrastructure to get at the assembly -- the code in the module tests to load an assembly and do tests seemed to do more than required by way of repeating all symbol reading combinations.

jbevain:
Thank you!

PR #711, opened 2020-12-29
Allow removing value of ModuleDefinition::EntryPoint

marek-safar:


marek-safar:
@jbevain PTAL

Mono.Cecil/ModuleDefinition.cs
```diff
@@ -539,15 +540,18 @@ public sealed class ModuleDefinition : ModuleReference, ICustomAttributeProvider
 
 		public MethodDefinition EntryPoint {
 			get {
-				if (entry_point != null)
+				if (entry_point_set)
 					return entry_point;
 
 				if (HasImage)
 					return Read (ref entry_point, this, (_, reader) => reader.ReadEntryPoint ());
```
jbevain, line 547: We need to set `entry_point_set` to `true` upon reading it from metadata as well otherwise we're reading from metadata every time now.

jbevain:
Thank you Marek!

PR #713, opened 2021-01-26
Add ModuleDefinition.ImmediateRead

mrvoorhe:
We are going to use this to do a multi stage load in parallel.  We will be able to greatly reduce our load times using this new API.  The way it's going to work is

Stage 1 - In parallel, load the assemblies with ReadingMode.Deferred.
Stage 2 - Populate our AssemblyResolver with a cache of all read assemblies.
Stage 3 - In parallel, call ImmediateRead.

What I really want is an API to load everything in stage 3.  I found that ImmediateRead does not load method bodies.  I don't know if/how you want want something like this exposed via the public API.  For now I'm iterating the data model and forcing things to load that ImmediateRead did not cover.

### DIFF
--- a/Test/Mono.Cecil.Tests/ModuleTests.cs
+++ b/Test/Mono.Cecil.Tests/ModuleTests.cs
@@ -277,6 +277,27 @@ namespace Mono.Cecil.Tests {
 			}
 		}
 
+		
+		[Test]
+		public void OpenModuleDeferredAndThenPerformImmediateRead ()
+		{
+			using (var module = GetResourceModule ("hello.exe", ReadingMode.Deferred)) {
+				Assert.AreEqual (ReadingMode.Deferred, module.ReadingMode);
+				module.ImmediateRead ();
+				Assert.AreEqual (ReadingMode.Immediate, module.ReadingMode);
+			}
+		}
+		
+		[Test]
+		public void ImmediateReadDoesNothingForModuleWithNoImage ()
+		{
+			using (var module = new ModuleDefinition ()) {
+				var initialReadingMode = module.ReadingMode;
+				module.ImmediateRead ();
+				Assert.AreEqual (initialReadingMode, module.ReadingMode);
+			}
+		}
+
 		[Test]
 		public void OwnedStreamModuleFileName ()
 		{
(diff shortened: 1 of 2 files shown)


Mono.Cecil/AssemblyReader.cs
```diff
@@ -166,6 +166,7 @@ protected override void ReadModule ()
 
 		public void ReadModule (ModuleDefinition module, bool resolve_attributes)
 		{
+			module.ReadingMode = ReadingMode.Immediate;
```
mrvoorhe, line 169: @jbevain You didn't include this in the changes you sent to me.  I added this for 2 reasons.

1) It seemed like the easiest way to write a test that meant something.

2) I saw some other behaviors such as writing and symbol reading checked the reading mode.  It seemed like it would be better if the reading mode was updated.  However, I'm not certain this is correct.  Does this seem sensible to you?

mrvoorhe:
@jbevain Can you take a look?

mrvoorhe:
I updated the PR to more sensibly handle a module with no image.  Now ImmediateRead will do nothing if there was no image.  Previously it would throw.

mrvoorhe:
@jbevain Does this seem like a change you'd be willing to take?

Test/Mono.Cecil.Tests/ModuleTests.cs
```diff
@@ -277,6 +277,27 @@ public void OpenModuleDeferred ()
 			}
 		}
 
+		
+		[Test]
+		public void OpenModuleDeferredAndThenPerformImmediateRead ()
+		{
+			using (var module = GetResourceModule ("hello.exe", ReadingMode.Deferred)) {
+				Assert.AreEqual (ReadingMode.Deferred, module.ReadingMode);
+				module.ImmediateRead();
```
jbevain, line 286: Missing a space before the opening (

Mono.Cecil/AssemblyReader.cs
jbevain: I say just move this to `ImmediateRead ()`

jbevain:
Thanks Mike, please just address the two tidbits and that's good to go!

Test/Mono.Cecil.Tests/ModuleTests.cs
mrvoorhe: Fixed

Mono.Cecil/AssemblyReader.cs
mrvoorhe: Done

jbevain (approved):

jbevain:
Thanks Mike!

PR #720, opened 2021-02-09
revert layout change

mrvoorhe:
https://github.com/jbevain/cecil/pull/504 seems incorrect.  Roslyn does not add a table entry for value types with no instance fields.  The result of that PR is that opening an assembly with cecil, writing it back out, and then reading it in again results in different values returned from `TypeDefinition.PackingSize` and `TypeDefinition.ClassSize`.  Even if the new values of `0` and `1` are harmless, it doesn't seem like a good idea to have cecil deviate from Roslyn's behavior.


atykhyy:
I agree with @mrvoorhe's conclusion, it was not the correct thing to do.

PR #721, opened 2021-02-09
Fix reading broken MDBs emitted by mcs. If sequence point ends at row…

mrvoorhe:
… -1, assume that the sequence point is just for a single line.


mrvoorhe:
Test failure fixed.  The expected result has been updated.  Now the expected results match the expected results of the `PdbTests`.  The `-1` in the MdbTests was always wrong.

jbevain:
Shouldn't this be dealt with client side?

If we're reading `6:-1:-1:-1`, we're going to write back `6:6:-1:-1` no?

mrvoorhe:
I'll check with @TautvydasZilys who originally made the change and get back to you.  Thanks for taking a look.

mrvoorhe:
@jbevain that's a good point.  After talking with @TautvydasZilys we are just going to drop this change.  We don't use mcs anymore so we'll be fine without this fix going forward.

Issue #652, opened 2020-02-21
ModuleDefinition.Assembly is null in multi-module assemblies


sajagi:
I am not sure whether this qualifies as a bug or is it intentional. `ModuleDefinition.Assembly` property in `AssemblyDefinition.Modules` is `null` for modules which are not defined in the same dll as the assembly. Given that the the assembly definition always creates new instances of module definitions it would make sense to keep the link back to the assembly. 

```csharp
public void Test()
{
    var module = ModuleDefinition.ReadModule(@"c:\Windows\Microsoft.NET\Framework64\v4.0.30319\System.EnterpriseServices.dll");
    
    foreach (var m in module.Assembly.Modules)
        Console.WriteLine($"{m.Name} assembly: {m.Assembly?.Name.Name ?? "null"}");

    // prints: 
    // System.EnterpriseServices.dll assembly: System.EnterpriseServices
    // System.EnterpriseServices.Wrapper.dll assembly: null
}
```

The problem with current situation is that currently there is no way to associate TypeDefinitions in such auxiliary modules back to the assembly they were defined in. 

Thanks for your hard work!

jbevain:
I think the original intent was that there's no Assembly row in a multi-module assembly.
So in theory, a multi module assembly could even belong to 2 assemblies.
Also I like to pretend that multi-module don't really exist :)

jbevain:
I think the way to go would be:

If you use `AssemblyDefinition.ReadAssembly` on a multi-module assembly, the `.Assembly` property should return the main assembly.
If you use `ModuleDefinition.ReadModule` on a `.netmodule`, the `.Assembly` property should return null.

jbevain:
Fixed by #730

Issue #725, opened 2021-03-11
NullReferenceException when trying to resolve ModuleReference members


AnakinSklavenwalker:
The testcode i used is copied from https://docs.microsoft.com/en-us/dotnet/framework/app-domains/build-multifile-assembly

### Scenario:
I compiled the tutorial from the link above. 
This yields the following files: `myAssembly.exe` `Client.netmodule` and `Stringer.netmodule` 
I'm trying to resolve as many Reference members into their matching Definitions (e.g. TypeReference --> TypeDefinition)

### Error Description:
i'm running the following code: 
```cs

 var appA = AssemblyDefinition.ReadAssembly("myAssembly.exe");

// Get all types from module 'Client.netmodule'
foreach (var t in modules_A.Skip(1).First().Types)
{
    if (t.Name == "<Module>")
        continue;

    // IL_0013: callvirt System.Void myStringer.Stringer::StringerMethod()
    var callInst = t.Methods.First(m => m.Name.Equals("Main")).Body.Instructions[7];

    if (callInst.Operand is not MethodReference methodReference)
        continue;

    // Type Name: myStringer.Stringer
    // TypeRef Scope: Stringer.netmodule ScopeType:ModuleReference
    var methodTypeRef = methodReference.DeclaringType;

    var methodTypeDef = methodTypeRef.Resolve();
}
```

The line `var methodTypeDef = methodTypeRef.Resolve();` fails with an NRE becuase: 
- `methodTypeRef `'s module is 'Client.netmodule'
- The ModuleDefinition of 'Client.netmodule' has no Assembly (`null`). According to ECMA this is valid.
- The MetadataResolver implementation is as follows: `var modules = type.Module.Assembly.Modules;` As you see it assumes, a ModuleDefinition always has an Assembly assigned to it. 

### Solutions
I see a couple of solutions here.
1. Check whether the module has an assembly
  a. return `null` in that case, indicating that it was impossible to find a proper TypeDefinition
  b.  Have some heuristic-based logic which tries to find a ModuleDefinition which matches the ModuleReferenc's name. Either lookup the assembly resolver's lookup directories for such a module or try on some cached ModuleDefinitions. Return null when no module was found or no module contained to TypeDefinition of the searched TypeReference. Or return the TypeDefinition if found. I think it should be noted however that this approach might return a wrong TypeDefinition if the wrong module matched.
2. Same as for 1. But instead of returning null, throw a proper exception.

[...]

PR #730, opened 2021-03-15
Fix resolving from a ModuleReference

AnakinSklavenwalker:
As requested here is the failing PR for #725 .
I managed to use the existing test resources `mma.exe` to reproduce the error. 

The test simple asserts that no exception shall get thrown, but since I'm not sure how you want to handle this situation, this is just a placeholder and should be replaced once a decision on actual behavior is made.

### DIFF
--- a/Mono.Cecil.PE/ImageWriter.cs
+++ b/Mono.Cecil.PE/ImageWriter.cs
@@ -784,7 +784,7 @@ namespace Mono.Cecil.PE {
 
 		int GetStrongNameLength ()
 		{
-			if (module.Assembly == null)
+			if (module.kind == ModuleKind.NetModule || module.Assembly == null)
 				return 0;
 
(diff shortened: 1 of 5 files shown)


jbevain:
That's related to #652. Funny that the code assumes this works.

jbevain:
Also fix #652

AnakinSklavenwalker:
changes look good for me, thanks. 
Maybe a few other thoughts:
1. I guess it should be NOT part of Cecil to find a weak "ModuleReference match". But since MetadataResolver::Resolve is `virtual` i guess there should be possibility for anyone to have custom logic. I like that! 
2. At this point I think it would be cool to have a nullable syntax/doc for properties. Are there any concerns about this? Is this a new issue? Enabling C# nullable syntax might reveal more such issues.

PR #733, opened 2021-03-15
Implement automatic handling of the HasFieldRVA field attribute.

### LINKED ISSUE #728: Mono.Cecil does not set the HasFieldRVA field on a FieldDefinition when the InitialData property is set to non-null
I'm trying to use the `FieldDefinition.InitialData` property to emit a static array into an image using the `.data` declaration feature in ECMA-335.

However, it seems that any data assigned to this field is dropped and not emitted as a `.data` declaration in the image.

I know that Mono.Cecil supports round-tripping the `InitialData` field from a loaded assembly back to disk if it hasn't been modified, but it seems that modifying it or assigning it on a new field is unsupported.

jkoritzinsky:
Fixes #728

### DIFF
--- a/Mono.Cecil/FieldDefinition.cs
+++ b/Mono.Cecil/FieldDefinition.cs
@@ -113,14 +113,8 @@ namespace Mono.Cecil {
 			}
 			set {
 				initial_value = value;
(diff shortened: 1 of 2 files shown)


jbevain:
Thank you for the PR @jkoritzinsky!

PR #734, opened 2021-03-16
Support custom scripts with file name info

simonferquel:
Many operations involved in rewriting debug symbols assume the rewritten
assembly base stream is a FileStream to get file name information. This
may not always be the case.

To cover this, this changes the way file name is resolved so that a
custom stream implementing `IHaveAFileName` can be understood by Cecil.

Signed-off-by: Simon Ferquel <simon.ferquel@unity3d.com>


sschoener:
To add to this: [This callsite](https://github.com/jbevain/cecil/blob/master/Mono.Cecil.Cil/PortablePdb.cs#L304) in particular will write out an empty PDB path unless you write out to a file. This is undesirable in scenarios where you modify the assembly in memory and write to a `MemoryStream` which is later copied to to a `FileStream`. The net effect is that the resulting assembly will contain an empty PDB path. (I'm not sure what the best solution to this is, but this is the problem we are seeing.)

jbevain:
Hey folks; Yeah I know that part is horrible.

Thanks for starting this discussion. I might not take this PR as is, but I'll think about how I want to fix this and get back to this PR.

Thanks again.

Issue #710, opened 2020-12-29
Is there a way to load debug symbols at runtime without any files


Serg046:
This is what I am trying to do:

```csharp
class Program
{
	static void Main(string[] args)
	{
		var asmDef = AssemblyDefinition.ReadAssembly(typeof(Test).Module.FullyQualifiedName,
			new ReaderParameters{ReadSymbols = true});
		asmDef.Name.Name += "Fake";
		var typeDef = asmDef.MainModule.GetType(typeof(Test).FullName, runtimeName: true).Resolve();
		var methodDef = typeDef.Methods.Single(m => m.Name == nameof(Test.Print));
		var consoleMethod = typeof(Console)
			.GetMethods()
			.Single(m => m.Name == nameof(Console.WriteLine) && m.GetParameters().Length == 1
			                                                 && m.GetParameters().Single().ParameterType ==
			                                                 typeof(string));
		methodDef.Body.Instructions.Insert(methodDef.Body.Instructions.Count - 1,
			Instruction.Create(OpCodes.Ldstr, "Injected"));
		methodDef.Body.Instructions.Insert(methodDef.Body.Instructions.Count - 1,
			Instruction.Create(OpCodes.Call, asmDef.MainModule.ImportReference(consoleMethod)));
		
		//This one works fine based on the pdb file
		var asm = WriteFileAssembly(asmDef);

		// But this one fails
		//var asm = WriteMemoryAssembly(asmDef);
		
		var type = asm.GetType(typeof(Test).FullName);
		type.InvokeMember(nameof(Test.Print), BindingFlags.InvokeMethod, null,
			Activator.CreateInstance(type), new object[0]);
	}

	private static Assembly WriteFileAssembly(AssemblyDefinition asmDef)
	{
		using var stream = File.Open("test.dll", FileMode.OpenOrCreate);
		asmDef.Write(stream, new WriterParameters {WriteSymbols = true});
		stream.Position = 0;
		// To avoid implicit load by the name
		File.Delete("renamed.pdb");
		File.Move("test.pdb", "renamed.pdb");
		// -----------------------------------
		var pdb = File.OpenRead("renamed.pdb");
		return AssemblyLoadContext.Default.LoadFromStream(stream, pdb);
	}

	private static Assembly WriteMemoryAssembly(AssemblyDefinition asmDef)
	{
		using var stream = new MemoryStream();
		using var symbolsStream = new MemoryStream();
		asmDef.Write(stream, new WriterParameters
		{
[...]

PR #735, opened 2021-03-17
Issues #710: Ability to load debug symbols from MemoryStream

Serg046:
I analyzed the whole file and seems that there is the only place where this issue could appear


jbevain:
As discussed in the issue, this is not perfect and I'll be looking at doing a better job of handling filenames, but this is good to go at the moment.

Thanks for your contribution @Serg046!

Issue #606, opened 2019-08-08
DocCommentId incorrect for a nested type of a generic class


joelmartinez:
Originating bug report on the .net core API docs:
https://github.com/dotnet/dotnet-api-docs/issues/2854

To summarize, we start with this method signature:
```csharp
public static ImmutableArray<TSource> ToImmutableArray<TSource>(this ImmutableArray<TSource>.Builder builder)
```

Compiler output is

```
M:System.Collections.Immutable.ImmutableArray.ToImmutableArray``1(System.Collections.Immutable.ImmutableArray{``0}.Builder)
```

and `DocCommentId` output is

```
M:System.Collections.Immutable.ImmutableArray.ToImmutableArray``1(System.Collections.Immutable.ImmutableArray`1.Builder{``0})
```

I verified it with the following minimal example:

```csharp
public class Test<T> {
	public class Builder {}

	public static Test<T> ToTest (Test<T>.Builder b) => new Test<T> ();
}
```
which gives the following result 

    M:Mono.Cecil.Tests.Test`1.ToTest(Mono.Cecil.Tests.Test`1.Builder{`0})

 Simple test case was written against the cecil master branch in the unit test project ...

jbevain:
@joelmartinez thanks for reporting this. Are you going to submit a PR for the issue?

joelmartinez:
@jbevain I will if it's not resolved by the time the original issue gets prioritized onto my plate ( ;) @mimisasouvanh). Not 100% sure when that will be, but I'll definitely need to resolve that at some point in the future ... so I'm definitely up for working on it at that point.

joelmartinez:
Ok, so doing some more investigation on this ... when I debug into the `DocCommentId` class, the `Builder` class's `.ToString()` method returns this:

```
Mono.Cecil.Tests.Test`1/Builder<T>
```

It is a `GenericInstanceType`, and `.HasGenericParameters == true`, but there are no elements in `.GenericParameters`, only one entry in `.GenericArguments` that contains the `T` parameter. Looking at [Roslyn's logic here](https://github.com/dotnet/roslyn/blob/master/src/Compilers/Core/Portable/DocumentationCommentId.cs#L548), it seems that `.IsGenericType` is false, based on the resulting docid in the docxml.

[...]

PR #749, opened 2021-04-21
Add support for generating the method and generic method comment signature with nested types

hiihellox10:
Hi @jbevain, current when a method or generic method with a nested type argument and that is defined in a generic class, the DocCommentId can't generate the comment signature of the method correctly. I created a few test methods with some special arguments for generate test XML documentation and make the DocCommentId can generate the same comment signature.

I don't have the specifications document of XML documentation that I now just found a [basic introduction](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/xmldoc/processing-the-xml-file) in Microsoft Docs but the introduction hasn't described the rules of the generic method. Although the comment signature of the generic method looks like has been fixed and maybe it's not a standard implementation. If you have the specifications document could you share them with me?

### DIFF
--- a/rocks/Mono.Cecil.Rocks/DocCommentId.cs
+++ b/rocks/Mono.Cecil.Rocks/DocCommentId.cs
@@ -145,22 +145,12 @@ namespace Mono.Cecil.Rocks {
 				throw new NotSupportedException ();
 
 			GenericTypeOptions options = new GenericTypeOptions {
-				StripGenericArity = true,
+				IsArgument = true,
 				IsNestedType = type.IsNested,
 				Arguments = type.GenericArguments
 			};
 
 			WriteTypeFullName (type.ElementType, options);
-			WriteGenericInstanceTypeArguments (type);
-		}
-
-		void WriteGenericInstanceTypeArguments (GenericInstanceType type)
-		{
-			if (!type.IsNested) {
-				id.Append ('{');
-				WriteList (type.GenericArguments, WriteTypeSignature);
-				id.Append ('}');
-			}
 		}
 
 		void WriteList<T> (IList<T> list, Action<T> action)
@@ -242,7 +232,7 @@ namespace Mono.Cecil.Rocks {
 
 			var name = type.Name;
 
-			if (options.StripGenericArity) {
+			if (options.IsArgument) {
 				var index = name.LastIndexOf ('`');
 				if (index > 0)
 					name = name.Substring (0, index);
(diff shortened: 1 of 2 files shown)


jbevain:
Hey @hiihellox10, thanks for the PR!
Hey @joelmartinez, I know you depend on this, could you review?

joelmartinez (approved):
Yes indeed ... this one was a long time coming. This looks great to me ... the added tests look like they match the generated IDs on the test classes.

@jbevain, it hasn't run the CI workflows. Not sure if this review will allow those to run, but if not I'd like to see the CI run, with your approval :)

joelmartinez:
Oh, also ... just for the sake of referencing, this is a fix to #606

saucecontrol:
> If you have the specifications document could you share them with me?

The spec Roslyn uses can be found [here](https://github.com/dotnet/csharplang/blob/main/spec/documentation-comments.md#processing-the-documentation-file), and the implementation is [here](http://sourceroslyn.io/#Microsoft.CodeAnalysis.CSharp/DocumentationComments/DocumentationCommentIDVisitor.cs) and [here](http://sourceroslyn.io/#Microsoft.CodeAnalysis.CSharp/DocumentationComments/DocumentationCommentIDVisitor.PartVisitor.cs).

Note that different compilers (e.g. Roslyn vs MSVC) and doc processing tools (e.g. whatever MS uses for the docs that ship with the framework) can differ in their documentation ID syntax, so there may be more than one "correct" ID for a given type/member.  ex: https://github.com/dotnet/roslyn/issues/52599

joelmartinez:
@saucecontrol "whatever ms uses for the docs" is [mdoc](https://github.com/mono/api-doc-tools) (which uses cecil's implementation). This is the reason that @hiihellox10 and I are working on this issue 🤣

Thanks for that link though, that was helpful. I _think_ that reading through it (albeit briefly) the changes here seem to match the specs.

saucecontrol:
Well that is *fascinating*.  I happened to find this PR while looking up the source for `GenericInstanceType` to make sure I was navigating it correctly while re-working my own doc tool to allow for multiple doc ID syntaxes to account for these differences.  Small world.  🤣

hiihellox10:
Hi @jbevain @joelmartinez, could you please help review this PR? Also, the CI is awaiting an approval.

rocks/Mono.Cecil.Rocks/DocCommentId.cs
```diff
@@ -211,13 +232,64 @@ void WriteTypeFullName (TypeReference type, bool stripGenericArity = false)
...
+
+				return typeGenericArguments;
+			}
+
+			return options.Arguments;
+		}
+
+		int GetGenericTypeParameterCount (TypeReference type)
+		{
+			var returnValue = 0;
```
jbevain, line 287: Doesn't `type.GenericParameters.Count` work here?

PR #755, opened 2021-05-15
Do not reset the stack size after a `break` instruction

ltrzesniewski:
This is just a minor correctness fix of something I noticed while reading the code.

`break` is not an unconditional jump instruction. It serves as a breakpoint for the debugger. This means it should not require the stack size of the next instruction to be zero (when there is no forward branch to it).

I don't think the C# compiler emits `break` instructions, so a unit test would have to use an assembly written in IL. I didn't think it's worth the trouble (as this change can only increase the `maxstack` value), but I can write one if you prefer.


jbevain:
That makes sense. Thank you!

PR #787, opened 2021-07-22
Update the version of Microsoft.NETFramework.ReferenceAssemblies.net40

tlakollo:
Update the version of Microsoft.NETFramework.ReferenceAssemblies.net40 since is detected as a package downgrade and a failure in newer SDKs

See failures in:
https://github.com/mono/linker/pull/2156/checks?check_run_id=3119342202

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -14,7 +14,7 @@
     <DisableImplicitFrameworkReferences>true</DisableImplicitFrameworkReferences>
   </PropertyGroup>
   <ItemGroup Condition=" '$(TargetFramework)' == 'net40' ">
-    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies.net40" Version="1.0.0" />
+    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies.net40" Version="1.0.2" />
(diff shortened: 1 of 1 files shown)


jbevain:
Thanks for the PR @tlakollo. Waiting for the checks to be green and this is good to go.

jbevain (approved):

agocke:
FYI I think this is because the SDK now automatically adds a reference to the appropriate ReferenceAssemblies package if the TF is netfx, but the package version is higher than the explicit reference that was used before, so it caused a NuGet warning.

jbevain:
@agocke @tlakollo do you know when this started? .NET 5? I'm not opposed to making the .NET 5 SDK the baseline to build Cecil and removing the explicit reference all-together and letting the SDK figure out what's best.

agocke:
It looks like auto-targeting was added 16 months ago, and looks new to the .NET 5 branches.
https://github.com/dotnet/sdk/commit/714c051d051fdf66a40c01a2d2c517a2a95dbfdb

The most recent break is just because of the package reference upgrade, so I think you could remove your reference entirely and require the .NET 5 SDK for building.

PR #806, opened 2021-10-29
Protect against circular type forwarders

### LINKED ISSUE #706: Defend against cyclical Type Forwarders?
This is not necessarily a bug in Cecil per se, but something to think about to protect consumers against cyclical type forwarders.

I'm filing this primarily so that there's a record of a lengthy investigation where the symptom was a StackOverflow in Cecil with this repeating stack:

```
Mono.Cecil.ExportedType.Resolve <--
Mono.Cecil.MetadataResolver.GetType
Mono.Cecil.MetadataResolver.Resolve
Mono.Cecil.ModuleDefinition.Resolve
Mono.Cecil.ExportedType.Resolve  <--
Mono.Cecil.MetadataResolver.GetType
Mono.Cecil.MetadataResolver.Resolve
Mono.Cecil.ModuleDefinition.Resolve
Mono.Cecil.TypeReference.Resolve
Mono.Cecil.TypeReference.ResolveDefinition
Mono.Cecil.MemberReference.Resolve
```

I'm attaching two .dlls where the type `System.Lazy``2` is type forwarded to the other .dll.

[CecilRepro.zip](https://github.com/jbevain/cecil/files/5610065/CecilRepro.zip)

Of course it's a bug where the custom assembly resolver I have resolved this combination of .dlls - normally there shouldn't be a cycle.

But I thought of how to defend against a StackOverflow here and there doesn't seem to be a way to defend against reentrancy on the consumer side. 

Ideally I'd place a circuit breaker here:
https://github.com/jbevain/cecil/blob/6edabe0525a17bda3613c2f4e5b28db568cb2c21/Mono.Cecil/ModuleDefinition.cs#L741

Remember the type I'm about to start resolving by adding it to the list, and remove it at the end of the method in a finally block. At the beginning of the method, check if the type is already in the list, and if so, throw an InvalidOperationException("recursive type references") or something.

I know it's kind of an obscure edge case but something to think about.

KirillOsenkov:
If we have two assemblies with type forwarders that point to each other, we enter an infinite loop and a stack overflow.

This breaks the cycle by detecting reentrancy.

Fixes https://github.com/jbevain/cecil/issues/706


KirillOsenkov:
@jbevain I even added a test!

jonlouie:
@jbevain What can community contributors do to get this PR merged and released? This resolves a bug that my team encountered.

KirillOsenkov:
FYI @jbevain adding the boolean field is free because of alignment padding:

image

KirillOsenkov:
Unfortunately it's not free for 32-bit (the size goes from 32-bit to 36 bit). But there are not many exported types per assembly, so it's honestly negligible.

```
Type layout for 'ExportedType'
Size: 36 bytes. Paddings: 3 bytes (%8 of empty space)
|==============================================|
| Object Header (4 bytes)                      |
|----------------------------------------------|
| Method Table Ptr (4 bytes)                   |
|==============================================|
|   0-3: String namespace (4 bytes)            |
|----------------------------------------------|
|   4-7: String name (4 bytes)                 |
|----------------------------------------------|
|  8-11: IMetadataScope scope (4 bytes)        |
|----------------------------------------------|
| 12-15: ModuleDefinition module (4 bytes)     |
|----------------------------------------------|
| 16-19: ExportedType declaring_type (4 bytes) |
|----------------------------------------------|
| 20-23: UInt32 attributes (4 bytes)           |
|----------------------------------------------|
| 24-27: Int32 identifier (4 bytes)            |
|----------------------------------------------|
|    28: Boolean reentrancyGuard (1 byte)      |
|----------------------------------------------|
| 29-31: padding (3 bytes)                     |
|----------------------------------------------|
| 32-35: MetadataToken token (4 bytes)         |
| |===============================|            |
| |   0-3: UInt32 token (4 bytes) |            |
| |===============================|            |
|==============================================|
```

cwensley:
We've been running into these same stack overflow issues quite a bit but intermittently and not related to circular type forwarders (from what I can tell).  I have confirmed this fixes our issues as well, so it'd be nice to have it included in an official release.

jbevain:
@KirillOsenkov just move the field with the other fields and this is good to go :)

KirillOsenkov:
It's happening.gif

mrvoorhe:
@jbevain @KirillOsenkov This change is breaking our multi-threaded immediate read logic.    Is this change worth it?  Cecil is generally a bit delicate when it comes to multi-thread but we've navigated successfully for years.  I could be mistaken, but I believe this change introduces a new gotcha pattern that didn't exist previously.

KirillOsenkov:
It is an unfortunate side effect that I've also discovered only recently, after the PR was merged.

The tradeoff here is the following: people used to run into stack overflows all the time without this check, and there's nothing you can do as a consumer to defend against that.

However you do have control over calling Resolve, so it's straightforward to just put a lock around that if you insist that you need parallelism.

We could also put a lock directly into the resolver to formalize the expectation that the reader is thread-safe. However the consumer could do that in the rare chance they need to.

This is my opinion only.

mrvoorhe:
> It is an unfortunate side effect that I've also discovered only recently, after the PR was merged.
> 
> The tradeoff here is the following: people used to run into stack overflows all the time without this check, and there's nothing you can do as a consumer to defend against that.
> 
> However you do have control over calling Resolve, so it's straightforward to just put a lock around that if you insist that you need parallelism.
> 
> We could also put a lock directly into the resolver to formalize the expectation that the reader is thread-safe. However the consumer could do that in the rare chance they need to.
> 
> This is my opinion only.

I can live with doing nothing if this change is preferred.  I was able to duplicate the Resolve logic into our code and avoid calling `ExportedType.Resolve` without too much trouble.

PR #807, opened 2021-10-29
Embed pdbs

KirillOsenkov:
What do you think?


jbevain:
I don't know, I like my DLL as small as possible for some reason :)

I'm not completely opposed to it though. Walk us through what do we gain?

KirillOsenkov:
Minor convenience for those debugging Cecil :)

But you decide. The .dll size increase is around 15-20%.

PR #808, opened 2021-11-01
Allow custom CoreLibs to be specified & custom 'ParameterDefinition' to supply 'index' argument

zezba9000:
I've been working on a project called IL2X again (a IL to C89, etc compiler): https://github.com/reignstudios/IL2X/tree/remake

First Mono.Cecil is missing a feature that allows me to specify non-standard CoreLib names. In this case "IL2X.CoreLib".
This is needed when compiling for non-standard .NET runtimes.

Second when IL2X resolves Generics for parameters it needs to create a new resolved 'ParameterDefinition' that uses the same index from the original unresolved param its based on (in short allow custom index arg in the constructor).

Third, same as ParameterDefinition but also for VariableDefinition.

### DIFF
--- a/Mono.Cecil.Cil/VariableDefinition.cs
+++ b/Mono.Cecil.Cil/VariableDefinition.cs
@@ -21,6 +21,12 @@ namespace Mono.Cecil.Cil {
 		{
 		}
 
+		public VariableDefinition (TypeReference variableType, int index)
+			: base (variableType)
+		{
+			this.index = index;
+		}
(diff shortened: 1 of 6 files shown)


jbevain:
Hi,

Thank you for your PR!

Unfortunately there are a multiple unrelated changes in there, with no tests and that seem useful only for your present use case.

For those scenarios you might be better off maintaining those changes on your end.

Thanks!

PR #819, opened 2021-12-13
Add importing of netmodule

### LINKED ISSUE #818: Create type in separate netmodule not supported
Take a look at this snippet
```
using Mono.Cecil;

var fooModule = ModuleDefinition.CreateModule("Foo", ModuleKind.NetModule);
var foo = new TypeDefinition("", "Foo", 0, fooModule.TypeSystem.Object);
fooModule.Types.Add(foo);

var mainModule = ModuleDefinition.CreateModule("Main", ModuleKind.NetModule);
mainModule.ModuleReferences.Add(fooModule);
var program = new TypeDefinition("", "Program", TypeAttributes.Class, mainModule.TypeSystem.Object);
program.Fields.Add(new FieldDefinition("foo", FieldAttributes.Static, mainModule.ImportReference(foo)));
mainModule.Types.Add(program);
```

Right now Cecil crashes with 
```
Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
   at Mono.Cecil.DefaultMetadataImporter.ImportScope(IMetadataScope scope)
   at Mono.Cecil.DefaultMetadataImporter.ImportScope(TypeReference type)
   at Mono.Cecil.DefaultMetadataImporter.ImportType(TypeReference type, ImportGenericContext context)
   at Mono.Cecil.DefaultMetadataImporter.ImportReference(TypeReference type, IGenericParameterProvider context)
   at Mono.Cecil.ModuleDefinition.ImportReference(TypeReference type, IGenericParameterProvider context)
   at Mono.Cecil.ModuleDefinition.ImportReference(TypeReference type)
```

This fails at this line https://github.com/jbevain/cecil/blob/7b8ee049a151204997eecf587c69acc2f67c8405/Mono.Cecil/Import.cs#L524
when attempt to get `Name` from `Assembly` which is not defined in this case.

So apparently Cecil not very well supports creation of netmodules.

kant2002:
Closes #818

### DIFF
--- a/Mono.Cecil/Import.cs
+++ b/Mono.Cecil/Import.cs
@@ -521,12 +521,31 @@ namespace Mono.Cecil {
 				return ImportReference ((AssemblyNameReference) scope);
 			case MetadataScopeType.ModuleDefinition:
 				if (scope == module) return scope;
-				return ImportReference (((ModuleDefinition) scope).Assembly.Name);
+				var moduleDefinition = (ModuleDefinition) scope;
+				if (moduleDefinition.Assembly == null)
+					return ImportReference (moduleDefinition);
+				else
+					return ImportReference (moduleDefinition.Assembly.Name);
 			case MetadataScopeType.ModuleReference:
-				throw new NotImplementedException ();
+					return ImportReference ((ModuleReference) scope);
 			}
 
 			throw new NotSupportedException ();
+		}
+
+		public virtual ModuleReference ImportReference (ModuleReference name)
+		{
+			Mixin.CheckName (name);
+
+			ModuleReference reference;
+			if (module.TryGetModuleReference (name, out reference))
+				return reference;
+
+			reference = new ModuleReference (name.Name);
+
+			module.ModuleReferences.Add (reference);
+
+			return reference;
 		}
 
 		public virtual AssemblyNameReference ImportReference (AssemblyNameReference name)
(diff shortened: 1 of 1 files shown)


jbevain:
Hi @kant2002! Thank you for your PR!

Is this an actual case you're hitting? While the code seems to make sense, could you add tests for it?

Thanks!

kant2002:
@jbevain I was hitting that in the actual snippet which I wrote in the issue. That snipped was created by @fornever when he play with C compiler in .NET. He probably can explain more.

Regarding tests, what's proper place to add test for this case? I was looking for them, but I cannot find nice place where I can validate case.

jbevain:
Hey!

The place to add a test would be there:

https://github.com/jbevain/cecil/blob/master/Test/Mono.Cecil.Tests/ImportCecilTests.cs

Thanks!

ForNeVeR:
> That snipped was created by @ForNeVeR when he play with C compiler in .NET. He probably can explain more.

This was an experiment while I was investigating how to create a `.netmodule` file using Cecil (in a project where I'm trying to create a fully C17-compatible compiler using Cecil, among other tools).

Initially, I wanted to mimic the full C-style compilation: creating a separate `.netmodule` for every translation unit, and then linking them together using the .NET assembly linker (`al.exe`, IIRC).

Though right now I've deserted this approach (not because of any Cecil problems, but because I don't think it's appropriate). So, there are no any actual uses of `.netmodule`s in my code at the moment.

It not impossible I'll decide to play with them again later, but not too likely, either.

kant2002:
@jbevain I think this is ready for another round of review.

kant2002:
@jbevain New year is soon. I behave well, so may I have some present on New Year, or on Christmas ?

PR #851, opened 2022-06-01
Fix mixed module ReadSymbols()

MarcoRossignoli:
If we try to read symbols for mixed module(native/CLI) we got this exception:

image

I noticed that in the same file there's already a "check" commented out(lines 49,51), we need only to be good for the read(sequence points) and we don't need to write mixed module(is not supported).

cc: @jakubch1 @fhnaseer


MarcoRossignoli:
@jbevain is it possible run tests from a Draft?

jbevain:
Thanks for the PR @MarcoRossignoli!

Issue #842, opened 2022-02-27
`MethodCallingConvention` lacks entry for "default unmanaged calling convention"


Zastai:
When using a function pointer in C#, the following syntax is available, with corresponding calling convention values:

```cs
  delegate* managed<type, ..., return-type> // same as leaving off the "managed"; value = 0
  delegate* unmanaged<type, ..., return-type> // means "use the default unmanaged calling convention for the platform; value 9
  delegate* unmanaged[Cdecl]<type, ..., return-type> // cdecl; value 1
  delegate* unmanaged[Fastcall]<type, ..., return-type> // cdecl; value 4
  delegate* unmanaged[Stdcall]<type, ..., return-type> // cdecl; value 2
  delegate* unmanaged[Thiscall]<type, ..., return-type> // cdecl; value 3
  delegate* unmanaged[xxx, yyy, ...]<type, ..., return-type> // also uses 9, and adds modopts to the return type for the listed conventions
```

So the "unmanaged without explicit calling convention" case currently does not have a name in the enum yet.

Suggested name: `DefaultUnmanaged` (I'd prefer `Default` with the current `Default` becoming `Managed`, because all values other than 0 imply unmanaged calls, but that's an API break).

(I can't find anything that suggests the enum values `VarArg` and `Generic` are usable in C# (perhaps they are a Mono-only thing?).)

Zastai:
Ah I see [here](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/function-pointers#metadata-representation-of-calling-conventions) that `default` is used in `CallKind` for the "managed" case, so that's fine. And it also lists `varargs` as specifically unsupported by C#.

So the entry needed is for what is called `unmanaged ext` in that section. `Unmanaged` or `DefaultIUnmanaged` seem like decent choices then.

Zastai:
Wish MS was better at documenting things properly in one place once a language proposal is accepted.

PR #852, opened 2022-06-08
Add `Unmanaged` calling convention

Zastai:
Fixes jbevain/cecil#842.


Zastai:
Would you take a version of the patch that adds XML doc comments to the enum? Would be really nice to work towards having the Cecil API documented, one step at a time.

jbevain:
Thanks for the PR!

PR #853, opened 2022-06-10
Emit sequence points in ascending offset order

### LINKED ISSUE #840: Problem adding sequence points to a method
Whenever I attempt to add a sequence point to a method with a portable PDB involved, the resulting PDB has a broken SequencePoints blob, resulting in

```
System.BadImageFormatException: Invalid compressed integer.
   at System.Reflection.Throw.InvalidCompressedInteger()
   at System.Reflection.Metadata.SequencePointCollection.Enumerator.MoveNext()
```

during processing, like when when hovering over the "SequencePoints" value when looking at the MethodDebugInformation table in ILSpy, or when running [pdb2xml](https://github.com/dotnet/symreader-converter).

Looking at a before/after, for a case where I added only a single "hidden" (lines set to 0xfeefee) sequence point, the original blob contained 376 bytes:
```
01-00-00-19-80-D8-07-05-00-1D-10-00-05-02-24-06
00-51-00-5D-18-00-25-00-4B-02-00-20-00-35-04-00
1F-00-3A-0A-00-16-00-00-81-2F-00-36-04-08-1F-00
34-02-00-1A-00-24-02-00-11-00-34-02-00-1A-00-24
02-00-11-00-06-02-00-01-00-36-04-00-1F-00-34-02
00-1A-00-24-02-00-11-00-34-02-00-1A-00-24-02-00
11-00-06-02-00-01-00-36-04-00-1F-00-34-02-00-1A
00-24-02-00-11-00-34-02-00-1A-00-24-02-00-11-00
06-02-00-01-00-36-04-00-1F-00-34-02-00-1A-00-24
02-00-11-00-34-02-00-1A-00-24-02-00-11-00-06-02
00-01-00-36-04-00-1F-00-34-02-00-1A-00-24-02-00
11-00-34-02-00-1A-00-24-02-00-11-00-34-02-00-1A
00-24-02-00-11-00-34-02-00-1A-00-24-02-00-11-00
34-02-00-1A-00-24-02-00-11-00-06-02-00-01-00-36
04-00-1F-00-34-02-00-1A-00-24-02-00-11-00-34-02
00-1A-00-3E-02-00-1F-00-24-02-00-11-00-06-02-00
01-00-36-04-00-1F-00-34-02-00-1A-00-24-02-00-11
00-34-02-00-1A-00-24-02-00-11-00-06-02-00-01-00
36-04-00-1F-00-34-02-00-1A-00-24-02-00-11-00-34
02-00-1A-00-24-02-00-11-00-06-02-00-01-00-36-04
00-1F-00-34-02-00-1A-00-24-02-00-11-00-34-02-00
1A-00-24-02-00-11-00-06-02-00-01-00-36-04-00-1F
00-34-02-00-1A-00-24-02-00-11-00-34-02-00-1A-00
24-02-00-11-00-01-0E-75
```
and the new, broken one contains 386 bytes:
```
01-00-00-19-80-D8-07-22-00-1D-10-00-06-02-24-06
00-51-00-5D-18-00-25-00-4B-02-00-20-00-35-04-00
1F-00-3A-0A-00-16-00-00-81-6A-00-36-04-08-1F-00
34-02-00-1A-00-24-02-00-11-00-34-02-00-1A-00-24
02-00-11-00-06-02-00-05-00-36-04-00-1F-00-34-02
00-1A-00-24-02-00-11-00-34-02-00-1A-00-24-02-00
11-00-06-02-00-05-00-36-04-00-1F-00-34-02-00-1A
00-24-02-00-11-00-34-02-00-1A-00-24-02-00-11-00
06-02-00-05-00-36-04-00-1F-00-34-02-00-1A-00-24
02-00-11-00-34-02-00-1A-00-24-02-00-11-00-06-02
00-05-00-36-04-00-1F-00-34-02-00-1A-00-24-02-00
11-00-34-02-00-1A-00-24-02-00-11-00-34-02-00-1A
00-24-02-00-11-00-34-02-00-1A-00-24-02-00-11-00
34-02-00-1A-00-24-02-00-11-00-06-02-00-05-00-36
04-00-1F-00-34-02-00-1A-00-24-02-00-11-00-34-02
00-1A-00-3E-02-00-1F-00-24-02-00-11-00-06-02-00
05-00-36-04-00-1F-00-34-02-00-1A-00-24-02-00-11
00-34-02-00-1A-00-24-02-00-11-00-06-02-00-05-00
36-04-00-1F-00-34-02-00-1A-00-24-02-00-11-00-34
02-00-1A-00-24-02-00-11-00-06-02-00-05-00-36-04
00-1F-00-34-02-00-1A-00-24-02-00-11-00-34-02-00
1A-00-24-02-00-11-00-06-02-00-05-00-36-04-00-1F
00-34-02-00-1A-00-24-02-00-11-00-34-02-00-1A-00
24-02-00-11-00-01-0E-75-FF-FF-F8-49-00-00-87-B9
00-00
```

My gaze is drawn by that FFFFF8 sequence near the end. The method in question includes a "hidden" sequence point (string hashing code for a string-based `switch`), so it can't be the 0xfeefee causing an issue there. My suspicion is that either the PDB writing code does not handle negative values correctly for a "compressed integer", or other processing results in a negative value when it shouldn't.

Looking at [the code that reads signed values back](https://github.com/dotnet/runtime/blob/976428fa4778faa3257b9cd7d6d3ea76a43410a0/src/libraries/System.Reflection.Metadata/src/System/Reflection/Metadata/BlobReader.cs#L490), the sign bit gets moved to the end. So if that 0xfffff849 is intended to be signed value -1975, I would expect to see something like 0x4f6f instead, if I read [the code](https://github.com/dotnet/runtime/blob/976428fa4778faa3257b9cd7d6d3ea76a43410a0/src/libraries/System.Reflection.Metadata/src/System/Reflection/Internal/Utilities/MemoryBlock.cs#L146) correctly (bit 0x40 in first byte to indicate: 2 bytes to read in total, then 0xf6f which is 1975 shifted left + the sign bit).

I'll try to make a minimal repro case, but hopefully this may already point in the right direction.

Zastai:
This is required by the Portable PDB specification, and the sequence
point collection in `MethodDebugInfo` does not guarantee it.

This also tweaks the writing of hidden sequence points to use two
writes of 0 as compressed unsigned integer, instead of a single
write of 0 as a plain short, to more closely match the specification.

Fixes #840.

Note: It would be extremely useful to get a 0.11.5 soon after this lands.

### DIFF
--- a/Mono.Cecil/AssemblyWriter.cs
+++ b/Mono.Cecil/AssemblyWriter.cs
@@ -3302,8 +3302,15 @@ namespace Mono.Cecil {
 			if (!info.TryGetUniqueDocument (out previous_document))
 				previous_document = null;
 
-			for (int i = 0; i < info.SequencePoints.Count; i++) {
-				var sequence_point = info.SequencePoints [i];
+			// The sequence points need to be emitted in strict ascending offset order, which is not guaranteed to be the case for
+			// info.SequencePoints.
+			var sequencePoints = new SortedDictionary<int, SequencePoint> ();
+			foreach (var sequence_point in info.SequencePoints) {
+				sequencePoints [sequence_point.Offset] = sequence_point;
+			}
+
+			var previous_offset = 0;
+			foreach (var sequence_point in sequencePoints.Values) {
 
 				var document = sequence_point.Document;
 				if (previous_document != document) {
@@ -3316,13 +3323,12 @@ namespace Mono.Cecil {
 					previous_document = document;
 				}
 
-				if (i > 0)
-					WriteCompressedUInt32 ((uint) (sequence_point.Offset - info.SequencePoints [i - 1].Offset));
-				else
-					WriteCompressedUInt32 ((uint) sequence_point.Offset);
+				WriteCompressedUInt32 ((uint) (sequence_point.Offset - previous_offset));
+				previous_offset = sequence_point.Offset;
 
 				if (sequence_point.IsHidden) {
-					WriteInt16 (0);
+					WriteCompressedUInt32 (0);
+					WriteCompressedUInt32 (0);
 					continue;
 				}
 


Zastai:
@jbevain Any chance of this landing?

jbevain:
@Zastai I think the assumption here is that it's on the user to provide a properly sorted SequencePoints collection.

Are there cases where we're reading portable pdbs and they end up unsorted in the collection?

Zastai:
On my phone at the moment (and it's late) so can't check right now, but I assume I couldn't easily get that collection sorted from the client side. Otherwise I expect I would have done that.

Zastai:
Ah I think I see the issue. My use case was flagging some code I add to a method as hidden, to make for smoother debugging. So I add a sequence point for the first of my newly-created insns.
But I don't think such a newly-added insn resolves to an actual offset (yet), so I am unclear on how I could reasonably sort the sequencepoints collection at that point.

I'm not sure it makes sense for Cecil to allow creation of debug data that violates the specification, so applying the sort as part of the writing seems like a reasonable thing to do.

If there is a safe way for me to do the sort myself (maybe preferably by adding a `Sort()` or `SortByOffset()` on the sequence point collection), then sure, Cecil doesn't have to sort it for me. But even then it would at least make sense to throw when an out-of-sequence entry is detected while writing (better to throw when trying to create a broken pdb than to break it and cause errors later).

Zastai:
@jbevain Any update on how you want me to proceed on this?

Zastai:
@jbevain Ping (and slightly disappointed this did not make it into 0.11.5).

Zastai:
@jbevain Another ping; it's now a year after this initial PR and it's still not clear how you want me to proceed.

I see three main options:
1. Accept this PR, ensuring that Cecil writes a vaild PDB when sequencepoints are added.
2. Change this PR to add sort functionality to the sequence points, e.g. adding a `Sort[ByOffset]()` extension method for `Collection<SequencePoint>`
  This assumes that this sort can be done safely (or at all) when newly created insns are referenced, because they might not have fully-resolved offsets yet. That would require clear documentation then that it should be called when finished manipulating a method body (e.g. after calling `Optimize()` on it.
3. Add a symbol writer option `SequencePointHandling`, with an enum of the same name, to choose the behaviour:
   a. `Keep` (the current behaviour): emit the sequence point in the order they are found in the collection, even when that results in an invalid PDB file
   b. `ThrowOnInvalidSequence`: same as `Keep`, but throw an exception if an invalid sequence is detected
   c. `EnsureValidSequence`: the behaviour of this PR: emit them in the correct sequence

leppie:
@jbevain I have tested fix as I started getting wrong info in portable PDB's converted from native.

WIth this fix you can roundtrip from native to portable and back to native and sequence points stay consistent. 

`PersistedAssemblyBuilder` in .NET 9 actually has the same bug when emitting a PDB.

leppie:
Some further investigation indicated that I was NOT creating sequence points out of order (that would be quite hard with ILGenerator), but were creating duplicate sequence points at the same offset. 

This is causing the problem.

PR #854, opened 2022-06-10
Add more style configuration

Zastai:
This sets up .NET naming for fields, locals and parameters to use
snake_case, so that editors won't complain about those.

It also disables trimming of trailing blanks to avoid accidentally
including such whitespace diffs in PRs.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,3 +9,15 @@ csharp_space_before_open_square_brackets = true
 csharp_new_line_before_open_brace = methods
 csharp_new_line_before_else = false
 csharp_indent_switch_labels = false
+
+trim_trailing_whitespace = false
+
+dotnet_naming_symbols.fields_locals_and_parameters.applicable_kinds = field, local, parameter
+dotnet_naming_symbols.fields_locals_and_parameters.applicable_accessibilities = *
+
+dotnet_naming_style.snake_case.capitalization = all_lower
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -8,4 +8,16 @@ csharp_space_between_method_call_name_and_opening_parenthesis = true
 csharp_space_before_open_square_brackets = true
 csharp_new_line_before_open_brace = methods
 csharp_new_line_before_else = false
-csharp_indent_switch_labels = false
\ No newline at end of file
+csharp_indent_switch_labels = false
+
+trim_trailing_whitespace = false
```
jbevain, line 13: That should be `true` no? We don't want trailing whitespaces.

SimonCropp: @jbevain i assume @Zastai was trying to define a config that most closely matches the current codebase. given there are existing trailing whitespaces, the assumption is `trim_trailing_whitespace = false`

Zastai: Exactly. I got a spurious diff for a PR because Rider had trimmed existing trailing white space.

If the intent is not to have trailing whitespace and it's ok for a PR to include such a change, then by all means set it to true.

jbevain: Ah got it, makes sense. Thanks. Yeah let's land this one like this, I'll definitely take a follow-up PR that trims trailing whitespaces.

jbevain:
Thank you!

PR #855, opened 2022-06-12
Add `MethodImplAttributes.AggressiveOptimization`

Zastai:
In [.NET Core 3.0 and later](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.methodimploptions?view=netcore-3.0), `MethodImplOptions` got a new flag value, `AggressiveOptimization` (512).

This adds that same flag to `MethodImplAttributes` (and a corresponding property to `MethodDefinition`). It also updates the comments for the flags to match the `MethodImplOptions` documentation better.


jbevain:
Thank you!

PR #687, opened 2020-08-25
Better way to update local scopes when method bodies are edited

vitek-karas:
This change will make sure that all local scopes are resolved (they refer to instructions directly and don't rely on instruction offset alone) then fixup local scopes by simply updating instruction references.

Added better tests to validate that the behavior is correct across PDB write/read and some variations of the resolved/unresolved scopes behavior.

PR #867, opened 2022-06-29
ILProcessor should also update custom debug info

vitek-karas:
When editing IL with ILProcessor various pieces of debug information have references to the origin IL instructions. These references can be either resolved (point to Instruction instance), in which case the editing mostly works, or unresolved (store IL offset only) in which case they need to be resolved before the editing can occur (after the edit the original IL offsets are invalid and unresolvable).

This is effectively a continuation of https://github.com/jbevain/cecil/pull/687 which implemented this for local scopes. This change extends this to async method stepping info and state machine scopes.

The change refactors the code to make it easier to reuse the same logic between the various debug infos being processed.

Updated the existing tests from https://github.com/jbevain/cecil/pull/687 to include async and state machine debug info (completely made up) and validate that it gets updated correctly.

This is an upstream port of https://github.com/mono/cecil/pull/34

### DIFF
--- a/Mono.Cecil.Cil/MethodBody.cs
+++ b/Mono.Cecil.Cil/MethodBody.cs
@@ -258,7 +258,7 @@ namespace Mono.Cecil.Cil {
 				item.next = current;
 			}
 
-			UpdateLocalScopes (null, null);
+			UpdateDebugInformation (null, null);
 		}
 
 		protected override void OnSet (Instruction item, int index)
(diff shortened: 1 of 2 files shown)


ltrzesniewski:
We noticed it after updating Cecil to f7b64f7, it didn't happen in the previous version that Fody used (a0a6ce4). It may be due to #810 or #824.

(sorry for the off-topic message)

Issue #856, opened 2022-06-14
Embedded PDBs not readable by VisualStudio


tom-englert:
With the latest changes 79b43e8 in `PortablePdb.cs` assemblies modified with cecil can no longer be debugged in VisualStudio when embedded pdbs are used. 

See https://github.com/Fody/Fody/issues/1110 for details how to reproduce.

jbevain:
:(

Thanks for the heads-up!

tom-englert:
I've created some test project without Fody, so this can be instantly checked within the cecil solution:
https://github.com/tom-englert/cecil/tree/%23856
![devenv_5EXsg5IJEt](https://user-images.githubusercontent.com/10229954/173795708-4e15d03e-4e35-4669-93c6-dbbae4e7e4f6.gif)
Just put a breakpoint on the unit test in the Testtee project and try to debug the test.

I've also reverted the changes to the PdbWriter to verify that this is the root cause:
https://github.com/tom-englert/cecil/tree/%23856_Fix

PR #869, opened 2022-07-14
Fix corrupted debug header directory entry when writing multiple such entries

vitek-karas:
Basically the first two entries are written correctly, and any after that which has data will have the RVA correct, but the virtual address field will be wrong. Depending on the consumer this can work (if they use RVA) or fail (if they use virtual address) as they would read garbage data.

Currently this mostly affects embedded protable PDBs since in that case we write 4 headers: CodeView, PdbChecksum, EmbeddedPdb and Deterministic (in this order), so the embedded PDB data is effectively wrong.

Also adds a test which validates that both the RVA and virtual address point to the same thing.

This fixes https://github.com/jbevain/cecil/issues/856 (I tried something very similar to the small repro posted in that issue)

/cc @tom-englert

### DIFF
--- a/Mono.Cecil.PE/ImageWriter.cs
+++ b/Mono.Cecil.PE/ImageWriter.cs
@@ -719,7 +719,7 @@ namespace Mono.Cecil.PE {
 					entry.Directory = directory;
 
 					data_len += entry.Data.Length;
-					data_address += data_len;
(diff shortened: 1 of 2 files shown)


tom-englert:
:+1:

jbevain:
Good catch, thanks @vitek-karas!

PR #870, opened 2022-08-12
InvariantCulture for operand to string conversion in Instruction.ToString()

Fantoom:
Instruction.ToString() uses the culture of the current thread, which can be problematic for converting doubles and floats to a string.

### DIFF
--- a/Mono.Cecil.Cil/Instruction.cs
+++ b/Mono.Cecil.Cil/Instruction.cs
@@ -9,6 +9,7 @@
 //
 
 using System;
+using System.Globalization;
 using System.Text;
 
 namespace Mono.Cecil.Cil {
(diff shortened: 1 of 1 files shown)


Mono.Cecil.Cil/Instruction.cs
```diff
@@ -122,11 +123,11 @@ public override string ToString ()
 				break;
 			case OperandType.InlineString:
 				instruction.Append ('\"');
-				instruction.Append (operand);
+				instruction.Append (Convert.ToString(operand, CultureInfo.InvariantCulture));
```
ForNeVeR, line 126: This is just a string, right? No need to convert a string into the invariant culture.

ForNeVeR (changes requested):

ForNeVeR:
For context: we're writing a C compiler using Cecil, and hit the issue while dumping the methods' IL code and comparing it in the compiler unit tests. So, I agree that these changes should be made.

Mono.Cecil.Cil/Instruction.cs
Fantoom: Did not notice that. Thanks!

ForNeVeR (approved):

jbevain:
Looking good, thank you!

PR #871, opened 2022-08-17
Add support for generic attributes

sbomer:
This adds support for generic attributes, added in C# 10. This is needed to fix a linker crash on runtime tests that used generic attributes: https://github.com/dotnet/linker/issues/2963


jbevain:
Looking good, thank you!

Issue #873, opened 2022-09-13
Constant type misidentified as primitive


SteveGilham:
The following code
```csharp
namespace NullConst {

    public class Program {
        public static void Main ()
        {
            new Program ().MakeConst ();
        }

        public void MakeConst ()
        {
            const IList<string> thing = null;
        }
    }
}
```
compiles to IL which does not contain the `const` value; but it is present in the debug metadata.

Reading and then immediately writing the assembly fails with an NRE claiming that there is a null primitive constant; when stepping through the process, the type information is correctly preserved in the debug metadata (still `IList<string>`, with element type `GenericInst`). the method `SignatureWriter GetConstantSignature (ElementType type, object value)`, however, sends this to the `default` primitive case where the null value fails.

The following reformulation of that method
```csharp
		private SignatureWriter GetConstantSignature (ElementType type, object value)
		{
			var signature = CreateSignatureWriter ();

			if (type == ElementType.String) {
				signature.WriteConstantString ((string)value);
			} else if (type.IsPrimitive ()) {
				signature.WriteConstantPrimitive (value);
			} else {
				signature.WriteInt32 (0);
			}

			return signature;
		}
```
passes the unit tests, and permits the assembly to be written.

PR #874, opened 2022-09-13
Address issue #873

SteveGilham:
This is the minimal change resolution to the issue, as outlined there.  

The test fell into no obvious existing category, so I put it in a new catch-all fixture.

The assembly NullConst.dll contains the source and project file used to build it as embedded resources.


jbevain:
Thanks!

PR #879, opened 2022-10-05
Fix a StackOverflowException reading windows runtime assemblies.

mrvoorhe:
During `AssemblyReader.ReadCustomAttributes` there is a call to `WindowsRuntimeProjections.Project`

```
if (module.IsWindowsMetadata ())
	foreach (var custom_attribute in custom_attributes)
		WindowsRuntimeProjections.Project (owner, custom_attributes, custom_attribute);
```

`WindowsRuntimeProjections.Project` would call `WindowsRuntimeProjections.HasAttribute`, which would then call `type.CustomAttributes`, which would end up back in `AssemblyReader.ReadCustomAttributes`. This would lead to a StackOverflowException.

This wasn't an issue previously.  My PR https://github.com/jbevain/cecil/pull/843 caused this sequence of calls to start resulting in a StackOverflowException.

Prior to my PR, there was a call to `metadata.RemoveCustomAttributeRange (owner);` before the call to `WindowsRuntimeProjections.Project`.  This meant that when `WindowsRuntimeProjections.HasAttribute` would call `type.CustomAttributes`, we'd still end up in `AssemblyReader.ReadCustomAttributes`, however, no attributes would be found because the following if would be true and lead to returning an empty collection.
```
if (!metadata.TryGetCustomAttributeRanges (owner, out ranges))
    return new Collection<CustomAttribute> ();
```

The old behavior was probably the wrong.  Although I'm not certain what the tangible impact was.

The fix was pretty easy.  `AssemblyReader.ReadCustomAttributes` will now pass in the custom attributes to `WindowsRuntimeProjections.Project` avoiding the need to call `type.CustomAttributes`

### DIFF
--- a/Test/Mono.Cecil.Tests/ImageReadTests.cs
+++ b/Test/Mono.Cecil.Tests/ImageReadTests.cs
@@ -182,7 +182,7 @@ namespace Mono.Cecil.Tests {
 		[Test]
 		public void WindowsRuntimeComponentAssembly ()
 		{
-			var resolver = WindowsRuntimeAssemblyResolver.CreateInstance ();
+			var resolver = WindowsRuntimeAssemblyResolver.CreateInstance (applyWindowsRuntimeProjections: false);
 			if (resolver == null)
 				return;
 
--- a/Test/Mono.Cecil.Tests/WindowsRuntimeProjectionsTests.cs
+++ b/Test/Mono.Cecil.Tests/WindowsRuntimeProjectionsTests.cs
@@ -17,19 +17,21 @@ namespace Mono.Cecil.Tests {
 		protected abstract string [] ManagedClassTypeNames { get; }
 		protected abstract string [] CustomListTypeNames { get; }
 
-		[Test]
-		public void CanReadMetadataType ()
+		[TestCase (true)]
+		[TestCase (false)]
+		public void CanReadMetadataType (bool readSdkAssembliesWithApplyWindowsRuntimeProjections)
 		{
 			if (Platform.OnMono)
 				return;
 
 			TestModule (ModuleName, (module) => {
 				Assert.AreEqual (ExpectedMetadataKind, module.MetadataKind);
-			}, verify: false, assemblyResolver: WindowsRuntimeAssemblyResolver.CreateInstance (), applyWindowsRuntimeProjections: true);
+			}, verify: false, assemblyResolver: WindowsRuntimeAssemblyResolver.CreateInstance (readSdkAssembliesWithApplyWindowsRuntimeProjections), applyWindowsRuntimeProjections: true);
 		}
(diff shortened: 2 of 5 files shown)


SimonCropp:
including a test that verifies the change might be helpful

mrvoorhe:
Fair point.  I'll see if I can come up with something

mrvoorhe:
I updated the PR with tests.  A few comments on the tests

* There was some logic to look in the 8.1 SDK.  This is outdated and shouldn't exist anymore.  I added logic to locate `Windows.winmd` from the windows 10 SDK.

* In order for the StackOverflowException to be triggered, `Windows.winmd` needed to be read with `ApplyWindowsRuntimeProjections=true`.  It turns out none of the windows runtime tests were doing that today.  I spoke with @TautvydasZilys about this and the conclusion was that it wouldn't be a bad thing if all of the windows runtime tests ran with `ApplyWindowsRuntimeProjections` set to `false` and to `true`.    That is what the `[TestCase(true)]` and `[TestCase(false)]` changes are for.

@jbevain please take a look.

Test/Mono.Cecil.Tests/ImageReadTests.cs
```diff
@@ -182,7 +182,7 @@ public void MetroAssembly ()
 		[Test]
 		public void WindowsRuntimeComponentAssembly ()
 		{
-			var resolver = WindowsRuntimeAssemblyResolver.CreateInstance ();
+			var resolver = WindowsRuntimeAssemblyResolver.CreateInstance (false);
```
jbevain, line 185: `applyWindowsRuntimeProjections: false`

Test/Mono.Cecil.Tests/WindowsRuntimeProjectionsTests.cs
```diff
@@ -17,19 +17,21 @@ public abstract class BaseWindowsRuntimeProjectionsTests : BaseTestFixture {
 		protected abstract string [] ManagedClassTypeNames { get; }
 		protected abstract string [] CustomListTypeNames { get; }
 
-		[Test]
-		public void CanReadMetadataType ()
+		[TestCase(true)]
```
jbevain, line 20: Style:

```csharp
[TestCase (true)]
[TestCase (false)]
```

jbevain:
Hi @mrvoorhe,

Thanks for the PR. Please address the tidbits and this is good to go.

Test/Mono.Cecil.Tests/ImageReadTests.cs
mrvoorhe: Fixed

mrvoorhe:
@jbevain I addressed your comments.  Thanks!

PR #882, opened 2022-11-01
Treat instance and static methods as different methods during resolution

joshpeterson:
When all other comparisons (return type, parameters, etc.) are the same, treat instance and static methods with the same name as different methods.

This corrects a problem we noticed in IL2CPP in Unity.


joshpeterson:
cc @jbevain

PR #888, opened 2022-12-27
Fix RVA field alignment

sbomer:
Upstreaming the change from https://github.com/dotnet/cecil/pull/55.

Fixes https://github.com/dotnet/runtime/issues/79477, specifically, the problem described in https://github.com/dotnet/runtime/issues/79477#issuecomment-1354917614.

This ensures that section starts are aligned by adjusting the previous `Range`'s length to ensure the computed start of a new `Range` meets the alignment requirements. It was done this way rather than just computing an aligned start for the new `Range`, because the `TextMap` assumes that the `Range`s are contiguous - see for example `GetNextRVA`.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,7 +12,7 @@ on:
 
 jobs:
   windows:
-    runs-on: windows-2019
+    runs-on: windows-latest
     steps:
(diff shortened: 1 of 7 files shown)


jbevain:
Thanks @sbomer!

PR #889, opened 2022-12-29
Make it able to save constants even if the enum is not resolved (jbevain#886)

Unkorunk:


teo-tsirpanis:
Duplicate of #702?

Unkorunk:
Not sure. My changes target to add possible save enum even if base types like int can't be resolved, because I think in current version of mono.cecil exist unnecessary extra check for user protection. I think it's good say to users if users do something wrong when it's possible, but approch when we block saving assembly only because user maybe did something wrong maybe didn't, i think it's not good approch.

p.s. this is my subjective opinion. I hope no one gets offended

teo-tsirpanis:
Both PRs were opened to fix the same bug; the difference is that #702 is more comprehensive and entirely avoids resolving types, while this one provides a fallback in case it was not found.

ForNeVeR:
@jbevain, could you please share your thoughts? I believe @teo-tsirpanis's summary above is correct. While you've marked #702 as an unacceptable trade-off, what about this fallback?

teo-tsirpanis:
> you've marked #702 as an unacceptable trade-off

> It also doesn't mean that it's not an acceptable trade-off.

He didn't say that, double negation can be confusing. 😅

ForNeVeR:
Ah, I'm sorry: I was, indeed, confused by the words there.

PR #902, opened 2023-04-06
Use v1.0.3 of the .NET FW reference assemblies

joshpeterson:
With MSBuild 7.0.201 and later, building the Mono.Cecil projects used by Unity's IL2CPP causes warning NU1605 to occur, indicating that there is a package version mismatch between the .NET Framework reference assemblies package used. To avoid this, bump the version used by Mono.Cecil and similar project to be the latest, version 1.0.3.

Note that MSBuild version 7.0.100 and earlier did not cause this warning to occur.


joshpeterson:
@jbevain Is this something that can be merged? I don't think it should have any negative impact.

jbevain (approved):

PR #916, opened 2023-06-29
Fix privatescope method resolution bug

mrvoorhe:
When resolving a `GenericInstanceMethod` for a `privatescope` method that has the same signature as other methods, `MetadataResolver.GetMethod` would incorrectly return the first method with the same name.

The fix for this seems to be an optimization opportunity as well.  Although I have to admit it feels a little too easy.  Please make sure I'm not overlooking some reason why this fix is not safe.

I added 2 variations of tests.

`PrivateScope` - These tests were fine and passed as is.  This is because the instruction operands are MethodDefinitions and therefore didn't need to be resolved by `MetadataResolver`

`PrivateScopeGeneric` - This test triggered the bug.

### DIFF
--- a/Test/Mono.Cecil.Tests/MethodTests.cs
+++ b/Test/Mono.Cecil.Tests/MethodTests.cs
@@ -2,6 +2,7 @@ using System;
 using System.Linq;
 
 using Mono.Cecil;
+using Mono.Cecil.Cil;
 using Mono.Cecil.Metadata;
 using Mono.Collections.Generic;
 using NUnit.Framework;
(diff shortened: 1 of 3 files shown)


Test/Mono.Cecil.Tests/MethodTests.cs
```diff
@@ -281,5 +282,93 @@ public void FunctionPointerArgumentOverload ()
...
+				// This is the first privatescope method.
+				var second_call_resolved = ((MethodReference)call_instructions [1].Operand).Resolve();
+				var expected_second_call_resolved = foo.Methods [first_same_name_index + 1];
+				
+				// Sanity check to make sure the ordering assumptions were correct.
+				Assert.IsTrue(expected_second_call_resolved.IsCompilerControlled, "The expected method should have been compiler controlled.");
+				
+				// The equality failure isn't going to be very helpful since both methods will have the same ToString value,
+				// so before we assert equality, we'll assert that the method is compiler controlled because that is the key difference
+				Assert.IsTrue(second_call_resolved.IsCompilerControlled, "Expected the method reference to resolve to a compiler controlled method");
```
mrvoorhe, line 313: This is the first assertion that would fail.  It would fail because `second_call_resolved` incorrectly resolved to `first_call_resolved`

jbevain (approved):

jbevain:
We keep learning interesting things. Thank you for the PR!

PR #939, opened 2024-03-23
Address issue 938 by choosing a FIPS compliant implementation

SteveGilham:
Replacing use of `SHA1Managed` with the compliant `SHA1CryptoServiceProvider` implementation.

Also, as a convenience, adding RollForward Major to the test projects so they will run on a machine with only net60 and up installed.


jbevain:
Thanks, I've fixed it in #944!

PR #948, opened 2024-06-04
Preserve custom debug information on types

sbomer:
Upstreaming change from https://github.com/dotnet/cecil/pull/185:

> Portable PDBs may have CustomDebugInformation on many metadata entities (see HasCustomDebugInformation in https://github.com/dotnet/runtime/blob/main/docs/design/specs/PortablePdb-Metadata.md#customdebuginformation-table-0x37).
> 
> For a select few of the CustomDebugInformation kinds (for example state machine hoisted local scopes), cecil has dedicated types to represent these in the object model. For the rest, cecil just reads the custom debug info out of the blob heap as a `byte[]`.
> 
> When writing back a module, cecil walks the metadata as represented in its object model, building the debug information as it goes.
> 
> So to support custom debug information for a new metadata token type:
> - the corresponding cecil type should be made to implement `ICustomDebugInformationProvider`,
> - the backing data for these getters should be populated on-demand, and also by the top-down walk in the immediate module reader, and
> - the top-down walk of metadata for writing should visit the cecil object and write its debug information.
> 
> This change implements the above for `TypeDefinition`. It extends the `ISymbolReader`/`ISymbolWriter` interfaces with a new method for reading/writing custom debug info for any `ICustomDebugInformationProvider`, and provides helpers for calling the symbol reader that can be used when adding custom debug information to other types in the future.
> 
> It doesn't include support for decoding the `TypeDefinitionDocument` debug info - it continues representing these as `BinaryCustomDebugInformation`.
> 
> Fixes https://github.com/dotnet/runtime/issues/100051.


jbevain:
Thanks!

PR #957, opened 2024-08-17
C#13: Add the "allow byref-like" anti-constraint

Zastai:
This adds a new `AllowsByRefLike` flag to `GenericParameterAttributes` which, unlike the others, is an anti-constraint (i.e. it removes restrictions rather than adding them) added in C# 13. Because of this, `AllowsByRefLike` was chosen as the corresponding bool property on `GenericParameter` (instead of
`HasAllowByRefLineConstraint`; the constant also does not include `Constraint` in its name).

At the same time, it adds the `NoSpecialConstraint` flag and a corresponding `HasNoSpecialConstraint` property.

Reference: corresponding constants in the CLR code: https://github.com/dotnet/runtime/blob/9daa4b41eb9f157e79eaf05e2f7451c9c8f6dbdc/src/coreclr/inc/corhdr.h#L850

### DIFF
--- a/Mono.Cecil/GenericParameter.cs
+++ b/Mono.Cecil/GenericParameter.cs
@@ -150,6 +150,11 @@ namespace Mono.Cecil {
 			set { attributes = attributes.SetMaskedAttributes ((ushort) GenericParameterAttributes.VarianceMask, (ushort) GenericParameterAttributes.Contravariant, value); }
 		}
 
+		public bool HasNoSpecialConstraint {
+			get { return attributes.GetMaskedAttributes ((ushort) GenericParameterAttributes.SpecialConstraintMask, (ushort) GenericParameterAttributes.NoSpecialConstraint); }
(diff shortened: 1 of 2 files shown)


Zastai:
Hmm, looks like it's already there; could have sworn I updated my fork.

Zastai:
Right, still keeping this PR because:
- it updates `SpecialConstraintMask` to include the new flag
- I kept the enum constant name in like with the CLR/COR header (no `Constraint` at the end, because it's not a constraint)
- the property name lacked the `Has` (needed because the `Allow` referred only to the `ByRefLike`, not the `Constraint`
  - I'm still keeping `AllowsByRefLike` for now, because it reads better than `HasAllowByRefLike`
    - but I'm open to changing it to the latter for consistency with the other property names
- it also adds the `NoSpecialConstraint` flag and property, to better match the whole of the CLR/COR version of the attributes

Issue #708, opened 2020-12-22
Reading some assemblies take a lot of memory (> 100x more than System.Reflection.Metadata)


joelverhagen:
I was doing some analysis of all assemblies on NuGet.org using Azure Functions and I found some packages failed with 
`OutOfMemoryException`. Many assemblies load without a problem, but I found several that take more than 1 GB of memory to load. The assembly itself is not even 5 MB so I'm not sure what's going on. Perhaps the DLL is obfuscated? No clue.

Anyway, I am trying to get the assembly public key token. Here's some example code, showing both memory consumption and durations.

```csharp
var path = "Oracle.ManagedDataAccess.dll";
var sw = Stopwatch.StartNew();
double mb = Math.Pow(1024, 2);
Console.WriteLine(Process.GetCurrentProcess().PrivateMemorySize64 / mb + " MB");

Console.WriteLine("== System.Reflection.Metadata ==");
using (var fileStream = File.OpenRead(path))
using (var peReader = new PEReader(fileStream))
{
    Console.WriteLine(peReader.GetMetadataReader().GetAssemblyDefinition().GetAssemblyName());
    Console.WriteLine(Process.GetCurrentProcess().PrivateMemorySize64 / mb + " MB");
    Console.WriteLine(sw.Elapsed);
}

Console.WriteLine("== Mono.Cecil ==");
using (var cecilModule = Mono.Cecil.ModuleDefinition.ReadModule(path))
{
    Console.WriteLine(cecilModule.Assembly.FullName);
    Console.WriteLine(Process.GetCurrentProcess().PrivateMemorySize64 / mb + " MB");
    Console.WriteLine(sw.Elapsed);
}
```

The output is:

<pre>
7.6640625 MB
== System.Reflection.Metadata ==
Oracle.ManagedDataAccess, Version=4.122.1.0, PublicKeyToken=89b483f429c47342
8.43359375 MB
00:00:00.0405080
== Mono.Cecil ==
Oracle.ManagedDataAccess, Version=4.122.1.0, Culture=neutral, PublicKeyToken=89b483f429c47342
<b>1601.359375 MB</b>
00:00:01.5745009
</pre>

Nearly 1.6 GB of memory to load this DLL. The DLL was found in this package: https://www.nuget.org/packages/Flexygo/1.0.13.36-rc

A minimal repro is here: [ConsoleApp2.zip](https://github.com/jbevain/cecil/files/5727621/ConsoleApp2.zip)

[...]

PR #963, opened 2024-12-05
Obfuscated file causes System.OutOfMemoryException

marci4:
Loading an obfuscated file (obfuscated with RedGate SmartAssembly 6) results in an out of memory exception because the SizeOfData in the DebugHeader is invalid

According to https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#the-debug-section the field 'Characteristics' is "reserved" and "must be zero".

Therefore ignore everything else if Characteristics != 0


marci4:
Could be a fix for #708
The behavior changed between 0.9.6.4 and 0.10 and I tracked it back to the support for debug header #349

PR #965, opened 2025-02-18
Preserve Documents debug info

### LINKED ISSUE #839: Assembly roundtrip discards documents from portable PDB
Compilation output will include all sources involved in the build as `Document` entries in the (portable) PDB.

This includes non-code sources, such as the `AssemblyInfo.cs` files generated by the build, for assembly/module attributes.

After loading an assembly with Mono.Cecil with symbol processing enabled and then writing it back, I observe that this has been reduced to a smaller set, presumably discarding any document that is not used by any sequence points.

I don't mind having methods like `DiscardUnusedDocuments()` available to "trim" debug info if I want to do so, but I would not expect that to happen by default. This discards potentially useful information.

### LINKED ISSUE #111138: SourceLink returning incorrect PDB info
**Version Used**: 4.13.0-2.24561.3 (24e1dd6b)

**Steps to Reproduce**:

1. Paste this code into a .NET 9 template:
```cs
IEnumerable<string>? s = null;
```
2. Make sure `dotnet.navigation.navigateToSourceLinkAndEmbeddedSources` is enabled in vscode. Or, if using vs, make sure sourcelink is enabled there.
2. Go-to-def on `IEnumerable`

**Expected Behavior**:

SourceLink opens https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/IEnumerable.cs

**Actual Behavior**:

SourceLink opens https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/TypeUnloadedException.cs.

sbomer:
Upstreaming the change from https://github.com/dotnet/cecil/pull/270.

Currently when writing PDBs, Document debug info is only preserved when referenced from a sequence point. It should be preserved always, because Documents don't necessarily correspond to any sequence points (for example, C# code that contains an interface definition with no implementation).

This will preserve the Document debug info always.

Fixes https://github.com/dotnet/runtime/issues/112049
Fixes https://github.com/dotnet/runtime/issues/111138
Fixes https://github.com/jbevain/cecil/issues/839


jbevain:
Thanks!

PR #970, opened 2025-05-02
Avoid unnecessarily creating TypeReferences on import

mrvoorhe:
When importing a TypeSpecification, an element type or generic argument could be a TypeDefinition within the current module. When this happens, a new TypeReference does not need to be created.  This can lead to a TypeReference being added to the typeref table for a type that is already in the assembly.  This seems to hold together, although I don't think it's ideal.  And really my goal is to avoid failing this logic in the ILLink test framework https://github.com/dotnet/runtime/blob/cec44d6dff9de95421f199f65bbc80b8296da1c0/src/tools/illink/test/Mono.Linker.Tests/TestCasesRunner/ResultChecker.cs#L56

This fix superceeds https://github.com/jbevain/cecil/pull/138.  I've left that fix in since that API is exposed publically and others could be depending on it.

While I was adjusting this logic, I thought I would apply the same logic to a TypeReference where the scope is the current modules assembly. This case is a bit contrived as it shouldn't happen, but if it did, it would result in the same circular reference problem as https://github.com/jbevain/cecil/pull/138

### DIFF
--- a/Mono.Cecil/Import.cs
+++ b/Mono.Cecil/Import.cs
@@ -506,6 +506,16 @@ namespace Mono.Cecil {
 			if (type.IsTypeSpecification ())
 				return ImportTypeSpecification (type, context);
 
+			// If a type's scope is the module definition then we don't need to create a new type reference.  We can reuse the existing one, which is likely the TypeDefinition.
+			// Reusing the type avoid creating an unnecessary entry in the type reference table
+			if (type.Scope == module)
+				return type;
+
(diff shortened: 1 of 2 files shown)


mrvoorhe:
@jbevain I don't think the linux CI failure is related to my changes.

samuraiGH:
> @jbevain I don't think the linux CI failure is related to my changes.

This is related to actions/runner-images#11101

I fixed it in #972

samuraiGH:
Now you can just sync your branch

mrvoorhe:
@jbevain What do you think about these changes?

